PR #51, opened 2026-04-01
feat: add h5i decision list --stale command

jonathanj502:
## Summary

Adds a `decisions` subcommand that queries recorded design decisions across commit history and flags those whose referenced code has drifted significantly since the decision was recorded.

- `h5i decisions list` — lists all decisions across history with commit OID, timestamp, location, choice, reason, and alternatives
- `h5i decisions list --stale` — filters to decisions where the referenced code has changed significantly, surfacing reasoning that may no longer apply
- `h5i decisions list --limit N` — controls how many commits to scan (default: 50)

## Motivation

The `--decisions` flag at commit time solves the write side of capturing design rationale. But recorded reasoning has a shelf life in that once the code it refers to changes significantly, the reasoning becomes incomplete and also potentially misleading. `decisions list --stale` fixes this by turning the decisions archive into an active signal that can be incorporated into `h5i resume` briefings alongside uncertainty signals and blind edits.

## Implementation

**New types in `metadata.rs`** (not serialized — computed at query time only):
- `StalenessStatus` — `Fresh` / `Stale { similarity }` / `Modified { similarity }` / `Unresolvable { reason }`
- `DecisionEntry` — decision + commit OID + timestamp + optional staleness

**Staleness detection** uses a two-tier strategy to keep the common case fast:
1. Compare ±5 lines of context around the recorded location. If unchanged → `Fresh`, done.
2. If context changed, compute AST similarity (0.0–1.0). Below 0.5 → `Stale`, at or above → `Modified`. Falls back to whole-file content comparison for file types without an AST parser.

**`decisions_list()`** in `H5iRepository` returns all entries unfiltered — display filtering is left to the caller so the method remains reusable for future MCP tool exposure.

## Design decisions

- `Decision.location` stays a free-form string; `FileLine` / `FileOnly` / `NonPath` classification happens at query time, not at record time — keeps the JSON schema simple and human-writable
- `decisions_list()` returns all entries; the CLI filters for `--stale` — avoids baking display logic into the data layer
- `StalenessStatus` and `DecisionEntry` do not derive `Serialize`/`Deserialize` — these types are never persisted

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,8 @@ h5i team status fix-auth                          # note the generated agent ids
 scripts/team-launch.sh fix-auth --task task.md
 
 # 3. Each agent peer-reviews, and revises inside its box
-scripts/team-review.sh fix-auth
+h5i team auto-peer-review fix-auth                # sync → freeze → mutual grant → instruct
+# (scripts/team-review.sh fix-auth is the shell equivalent, adds --relaunch)
 
 # 3. the neutral verdict: replay each candidate, run the tests, merge the winner
 # h5i team sync     fix-auth                       # ingest agents' staged work (no relaunch)
--- a/src/main.rs
+++ b/src/main.rs
@@ -2199,6 +2199,23 @@ enum TeamCommands {
         #[arg(long)]
         json: bool,
     },
+    /// Open the peer-review round in one shot: sync staged work, freeze the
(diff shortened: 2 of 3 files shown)


jonathanj502:
Rebased onto the latest main to resolve merge conflicts with #52 (vibe, policy, compliance commands). No logic changes to the --stale feature.

PR #53, opened 2026-04-03
feat: stale decision commit warnings

jonathanj502:
Directly extends decisions --stale feature. When `h5i commit` runs, check whether the staged diff touches a line near a previously recorded stale decision. If so, print an inline warning before the commit completes:

  ⚠  Commit touches src/repository.rs:88 near a stale decision from abc123: "chose JSON over binary"
     reason was:  reduces parse overhead at startup
     Has this reasoning changed? Run `h5i decisions --stale` for details.

## Design

- **Line-level matching, not file-level.** Uses diff hunk ranges from `git2::Patch` against the decision's recorded `file:line` location (±`CONTEXT_LINES` proximity). File-level matching produces too many false positives on large files.
- **Stale-only signal.** `Modified` and `Fresh` entries are dropped. Only decisions with similarity < 0.5 produce a warning.
- **Non-blocking.** Errors from the staleness check are swallowed so a bug here never breaks the commit path.

## Testing

Two unit tests added:
- `fires_when_line_touched` — hunk spanning the decision line triggers the warning
- `silent_when_far` — hunk far from the decision line produces no warning

### DIFF
--- a/roadmap/new-commands.md
+++ b/roadmap/new-commands.md
@@ -1,0 +1,54 @@
+## feat: add h5i decision list --stale command
+
+### Summary
+
+Adds a `decisions` subcommand that queries recorded design decisions across commit history and flags those whose referenced code has drifted significantly since the decision was recorded.
+
+- `h5i decisions list` — lists all decisions across history with commit OID, timestamp, location, choice, reason, and alternatives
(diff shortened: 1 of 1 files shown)


Koukyosyumei:
Thanks for the proposal and implementation! Right now, improving the other commands is a higher priority, but I’ll record this in the roadmap and merge it.

PR #266, opened 2026-07-06
feat(filters): add bun output filter

### LINKED ISSUE #206: Add an output filter for `bun`
## Add an output filter for `bun`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time. This is a self-contained good-first-issue: **add a filter for `bun`**. No changes to core logic are needed.

### Why
`bun install`/`bun test` print per-package and per-test progress; the signal is failures, the install summary, and counts.

### What to keep / strip
- **Keep (the signal):** failing tests, the `N pass / M fail` summary, and install error lines.
- **Strip (the noise):** per-package install spam and blank lines.

### Steps
1. Create `assets/filters/bun.toml`.
2. Match the command with a `match_command` regex, e.g. `match_command = "^bun\\b"`. Make it match typical invocations but not unrelated commands.
4. Add **2–3 inline golden tests** (`[[tests.bun]]`): a normal run, an error/failure run, and a clean/empty run if relevant.

### Golden-test shape
```toml
[filters.bun]
description = "…"
match_command = "^bun\\b"
strip_ansi = true

[[tests.bun]]
input = """…raw output…"""
```

### How to verify
```bash
```

### Reference
The upstream [`rtk`](https://github.com/rtk-ai/rtk) project supports the Bun toolchain — see its `src/filters/` and `src/cmds/` for inspiration on what matters in the output. h5i's filters are derived from rtk (see `assets/filters/NOTICE`).

### Acceptance criteria
- [ ] `assets/filters/bun.toml` added with a clear `description`
- [ ] `match_command` matches typical `bun` invocations and not unrelated ones

zerone0x:
## Summary
Fixes #206

Adds a declarative built-in output filter for `bun` so `h5i capture run` keeps test/install summaries and failure signal while dropping package/progress noise.

## Changes
- Add `assets/filters/bun.toml` with a `^bun\b` command matcher
- Keep Bun test pass/fail summaries, failure/error lines, and install summaries
- Add inline golden tests for passing tests, failing tests, install output, and empty output

## Test Plan

### DIFF
--- a/assets/filters/bun.toml
+++ b/assets/filters/bun.toml
@@ -1,0 +1,75 @@
+[filters.bun]
+description = "Compact Bun output — keep test/install summaries and failures, drop per-package/progress noise"
+match_command = "^bun\\b"
+strip_ansi = true
+keep_lines_matching = [
+  "(?i)\\b(error|failed|failure|fail|panic|exception)\\b",
+  "(?i)^\\s*(installed|saved lockfile|checked)\\b",
+  "(?i)^\\s*\\d+\\s+packages? installed\\b",
+  "^\\s*\\d+\\s+(pass|passed|fail|failed|skip|skipped|todo)s?\\b",
+  "^\\s*(✓|✗|×|error:)\\s+",
+]
+truncate_lines_at = 240
+max_lines = 100
+on_empty = "bun: clean"
+
+[[tests.bun]]
+name = "keeps test summary and drops passing-test noise"
+input = """
+bun test v1.2.17 (282dda62)
+
+tests/math.test.ts:
+(pass) add > sums two numbers [0.32ms]
+(pass) add > handles zero [0.08ms]
+
+ 2 pass
+ 0 fail
+ 4 expect() calls
+Ran 2 tests across 1 file. [7.00ms]
+"""
+expected = """ 2 pass
+ 0 fail"""
+
+[[tests.bun]]
+name = "keeps failing test details and summary"
+input = """
+bun test v1.2.17 (282dda62)
+
+tests/api.test.ts:
+(pass) api > healthcheck [0.20ms]
+(fail) api > returns user [1.04ms]
+  Error: expected 500 to be 200
+      at tests/api.test.ts:12:18
+
+ 1 pass
+ 1 fail
+ 3 expect() calls
+Ran 2 tests across 1 file. [8.00ms]
+"""
+expected = """(fail) api > returns user [1.04ms]
+  Error: expected 500 to be 200
+ 1 pass
+ 1 fail"""
+
+[[tests.bun]]
+name = "install keeps summary but drops package progress"
+input = """
+bun install v1.2.17 (282dda62)
+Resolving dependencies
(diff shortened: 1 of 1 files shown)


Koukyosyumei (changes requested):
## Review: bun output filter

Thanks for this! The mechanism is sound and all four golden cases pass (I traced each line-by-line through `apply_filter`, since the golden runner does an exact `actual.trim() == expected.trim()` compare). `RustEmbed` auto-embeds the file, `^bun\b` doesn't collide with `bundle-install` (`bun` + `d` has no word boundary), and the routing-ambiguity guard won't regress. A couple of design points before merge, though.

### Main concern: allowlist diverges from convention

This is a `keep_lines_matching` (allowlist) filter, but **68 of 69 existing filters use `strip_lines_matching` (denylist)** — only `spring-boot` uses keep. Every sibling package-manager/test filter (`npm`, `yarn`, `pnpm`, `jest`, `vitest`) is a denylist. The allowlist is riskier for a filter whose job is *preserving failure signal*:

- **Novel failures silently vanish.** Any failure that doesn't match the hardcoded vocabulary is dropped entirely — an uncaught promise rejection, a segfault/`Bus error`, a TypeScript/transpile error, or `console.error(...)` phrased without error/fail/panic/exception. A denylist lets these through by default.
- **Lost stack context.** In the "keeps failing test details" golden case, the frame `at tests/api.test.ts:12:18` is dropped — so a failure keeps `(fail) …` and `Error: …` but not *where* it failed, which is the most useful line for debugging.

Suggestion: flip to a `strip_lines_matching` denylist scoped to the known noise (passing `(pass)` lines, `Resolving/Resolved` progress, `+ pkg@ver` install lines) to match the sibling filters — or, if you keep the allowlist, add stack frames (`^\s+at\s`) and a generic uncaught-error catch.

### Secondary: `match_command` is too broad

`^bun\b` matches **every** bun invocation, including `bun run <script>` and `bun x`, whose arbitrary app/server output would then be run through the keep-allowlist and mostly discarded. Siblings scope to subcommands precisely to avoid this (`^pnpm\s+(install|i|add|run|…)`, `^npm\s+(install|ci|…)`). Recommend:

```toml
match_command = "^bun\\s+(test|install|i|add|remove|update|ci)\\b"
```

### Minor

- `\bfail\b` / `\berror\b` will keep a *passing* test whose title contains those words (`(pass) handles failure gracefully`) or a header like `tests/error-handling.test.ts:`. Harmless noise, but slightly at odds with the otherwise-aggressive drop.
- `strip_ansi = true` is correct and matches siblings. 👍

Mechanically correct and well-tested — requesting changes mainly to (1) scope `match_command` to bun's subcommands and (2) reconsider allowlist vs the denylist convention (or at least keep stack-trace lines so failures stay debuggable).

Koukyosyumei:

PR #267, opened 2026-07-06
feat(filters): add cmake output filter

### LINKED ISSUE #207: Add an output filter for `cmake`
## Add an output filter for `cmake`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time. This is a self-contained good-first-issue: **add a filter for `cmake`**. No changes to core logic are needed.

### Why
`cmake` configure output is mostly `-- Detecting/Found` probe lines; the signal is configuration errors and the final 'Configuring/Generating done' status.

### What to keep / strip
- **Keep (the signal):** `CMake Error`/`CMake Warning` blocks and the done/failed status.
- **Strip (the noise):** the `-- Check/Detecting/Found ...` probe lines and blank lines.

### Steps
1. Create `assets/filters/cmake.toml`.
2. Match the command with a `match_command` regex, e.g. `match_command = "\\bcmake\\b"`. Make it match typical invocations but not unrelated commands.
4. Add **2–3 inline golden tests** (`[[tests.cmake]]`): a normal run, an error/failure run, and a clean/empty run if relevant.

### Golden-test shape
```toml
[filters.cmake]
description = "…"
match_command = "\\bcmake\\b"
strip_ansi = true

[[tests.cmake]]
input = """…raw output…"""
```

### How to verify
```bash
```

### Reference
The upstream [`rtk`](https://github.com/rtk-ai/rtk) project supports build tooling — see its `src/filters/` and `src/cmds/` for inspiration on what matters in the output. h5i's filters are derived from rtk (see `assets/filters/NOTICE`).

### Acceptance criteria
- [ ] `assets/filters/cmake.toml` added with a clear `description`
- [ ] `match_command` matches typical `cmake` invocations and not unrelated ones

zerone0x:
## Summary
Fixes #207

Adds a declarative built-in output filter for `cmake` configure/build output, preserving warnings, errors, and final configure/generate status while dropping routine probe chatter.

## Changes
- Add `assets/filters/cmake.toml` with a `^cmake\b` matcher
- Strip common `-- Detecting` / `-- Found` / compiler probe lines and blanks
- Add inline golden tests for success, error, warning, and empty output

## Test Plan

### DIFF
--- a/assets/filters/cmake.toml
+++ b/assets/filters/cmake.toml
@@ -1,0 +1,72 @@
+[filters.cmake]
+description = "Compact CMake configure/build output — strip probe chatter, keep warnings, errors, and final status"
+match_command = "^cmake\\b"
+strip_ansi = true
+strip_lines_matching = [
+  "^\\s*$",
+  "^-- (The |Detecting |Check(ing)? |Looking for |Performing Test |Found |Using |Configuring incomplete, errors occurred!)",
+  "^-- (C|CXX|CUDA|ASM) compiler identification is ",
+  "^-- (Detecting|Check for working|Detecting .* done)",
+]
+truncate_lines_at = 240
+max_lines = 120
+on_empty = "cmake: clean"
+
+[[tests.cmake]]
+name = "configure success keeps final status and drops probes"
+input = """
+-- The C compiler identification is GNU 13.2.0
+-- The CXX compiler identification is GNU 13.2.0
+-- Detecting C compiler ABI info
+-- Detecting C compiler ABI info - done
+-- Check for working C compiler: /usr/bin/cc - skipped
+-- Found OpenSSL: /usr/lib/libssl.so
+-- Configuring done (0.4s)
+-- Generating done (0.1s)
+-- Build files have been written to: /work/build
+"""
+expected = """-- Configuring done (0.4s)
+-- Generating done (0.1s)
+-- Build files have been written to: /work/build"""
+
+[[tests.cmake]]
+name = "keeps error block while dropping probe noise"
+input = """
+-- The CXX compiler identification is GNU 13.2.0
+-- Detecting CXX compiler ABI info
+-- Detecting CXX compiler ABI info - done
+CMake Error at CMakeLists.txt:12 (find_package):
+  Could not find a package configuration file provided by "Qt6" with any of
+  the following names:
+
+    Qt6Config.cmake
+    qt6-config.cmake
+
+-- Configuring incomplete, errors occurred!
+"""
+expected = """CMake Error at CMakeLists.txt:12 (find_package):
+  Could not find a package configuration file provided by "Qt6" with any of
+  the following names:
+    Qt6Config.cmake
+    qt6-config.cmake"""
+
+[[tests.cmake]]
+name = "keeps warning and final status"
+input = """
+-- Detecting C compiler ABI info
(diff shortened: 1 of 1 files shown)


Koukyosyumei (approved):
## Summary

Adds a single new declarative filter asset, `assets/filters/cmake.toml`, matching `^cmake\b`. It strips CMake configure probe chatter while preserving warnings, errors, and the final configure/generate/build-files status, plus four inline golden tests. It's a clean, additive, self-contained change that follows the established pattern in `src/filter_rules.rs`.

## What's good

- **Correct engine usage.** All fields used (`description`, `match_command`, `strip_ansi`, `strip_lines_matching`, `truncate_lines_at`, `max_lines`, `on_empty`) are valid under `FilterDef`, which is `#[serde(deny_unknown_fields)]` (`src/filter_rules.rs:66`), so the file parses and `builtin_rules_all_parse` stays green.
- **Golden tests pass.** I traced all four cases through `apply_filter` (`src/filter_rules.rs:248`) and reproduced the engine (`.lines()` drops the trailing newline, `^\s*$` strips blanks, empty result → `on_empty`). All four PASS: success keeps final status, error/warning blocks are preserved, empty run collapses to `cmake: clean`. Good coverage of the meaningful states.
- **Preserves the important negatives.** `-- Could NOT find …` and `CMake Deprecation/Warning` lines do not match any strip pattern, so real signal is retained — exactly right.
- **Conventions match siblings.** Structure mirrors `make.toml`/`gcc.toml`; the em-dash in `description` is consistent with `gcc.toml`'s description, and `on_empty` mirrors `make: ok`. No manual registration needed (RustEmbed auto-includes `assets/filters/*.toml`), and there's no `match_command` overlap with any other rule, so routing stays unambiguous.

## Minor notes (non-blocking)

- **Redundant strip patterns.** Patterns 3 and 4 are largely subsumed by pattern 2:
  - `^-- (Detecting|Check for working|Detecting .* done)` — real CMake `Detecting …` lines always have a trailing space, already caught by pattern 2's `Detecting `, and `Check for working` is caught by pattern 2's `Check(ing)? `.
  - `^-- (C|CXX|CUDA|ASM) compiler identification is ` — real output is `-- The C compiler identification is …`, already caught by pattern 2's `The `.

  They work as harmless belt-and-suspenders (they'd catch hypothetical output lacking the `The `/space prefix), but consider trimming to keep the rule tight. Not required.
- **`-- Found <pkg>` is stripped.** This drops successful-find lines like `-- Found OpenSSL: … (found version …)`. That's a reasonable probe-chatter tradeoff since warnings/errors are kept, but worth being aware it also removes the version-confirmation some users scan for.
- **Build-phase interaction.** Since routing is first-match-per-command, `cmake --build .` is claimed by this rule, so compiler errors emitted during the build pass through without the `gcc` rule's include-chain stripping. Acceptable given the one-rule-per-command model; just noting the scope.
- **Test plan.** The author notes `cargo test builtin_golden_tests_pass` was blocked locally (no `cc`). I confirmed the four cases pass via a faithful simulation of the engine, but CI should be the source of truth here.

## Verdict

No blocking issues. Correct, well-tested, convention-consistent additive change. The regex-redundancy trim is optional polish and can be done in a follow-up or ignored.

PR #268, opened 2026-07-06
feat(filters): add ninja output filter

### LINKED ISSUE #208: Add an output filter for `ninja`
## Add an output filter for `ninja`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time. This is a self-contained good-first-issue: **add a filter for `ninja`**. No changes to core logic are needed.

### Why
`ninja` prints a `[n/total]` progress line per build step; the signal is compiler errors/warnings and the final result.

### What to keep / strip
- **Keep (the signal):** compiler error/warning lines and the failing target.
- **Strip (the noise):** the `[12/340] Building ...` progress lines and blank lines.

### Steps
1. Create `assets/filters/ninja.toml`.
2. Match the command with a `match_command` regex, e.g. `match_command = "^ninja\\b"`. Make it match typical invocations but not unrelated commands.
4. Add **2–3 inline golden tests** (`[[tests.ninja]]`): a normal run, an error/failure run, and a clean/empty run if relevant.

### Golden-test shape
```toml
[filters.ninja]
description = "…"
match_command = "^ninja\\b"
strip_ansi = true

[[tests.ninja]]
input = """…raw output…"""
```

### How to verify
```bash
```

### Reference
The upstream [`rtk`](https://github.com/rtk-ai/rtk) project supports build tooling (complements the C/C++ `gcc` filter) — see its `src/filters/` and `src/cmds/` for inspiration on what matters in the output. h5i's filters are derived from rtk (see `assets/filters/NOTICE`).

### Acceptance criteria
- [ ] `assets/filters/ninja.toml` added with a clear `description`
- [ ] `match_command` matches typical `ninja` invocations and not unrelated ones

zerone0x:
## Summary
Fixes #208

Adds a declarative built-in output filter for `ninja` output, preserving failed targets and compiler diagnostics while dropping routine `[n/total]` build-progress noise.

## Changes
- Add `assets/filters/ninja.toml` with a `^ninja\b` matcher
- Strip routine progress and blank lines
- Add inline golden tests for successful builds, failure diagnostics, warning output, and empty output

## Test Plan

### DIFF
--- a/assets/filters/ninja.toml
+++ b/assets/filters/ninja.toml
@@ -1,0 +1,59 @@
+[filters.ninja]
+description = "Compact Ninja output — strip routine build progress, keep failed targets and compiler diagnostics"
+match_command = "^ninja\\b"
+strip_ansi = true
+strip_lines_matching = [
+  "^\\s*$",
+  "^\\[\\d+/\\d+\\] (Building|Linking|Generating|Compiling|Scanning|Copying|Installing|Running|Creating|Updating)\\b",
+]
+truncate_lines_at = 240
+max_lines = 120
+on_empty = "ninja: ok"
+
+[[tests.ninja]]
+name = "successful build collapses progress noise"
+input = """
+[1/5] Building C object src/foo.o
+[2/5] Building CXX object src/bar.o
+[3/5] Linking CXX static library libapp.a
+[4/5] Building CXX object src/main.o
+[5/5] Linking CXX executable app
+"""
+expected = "ninja: ok"
+
+[[tests.ninja]]
+name = "keeps failed target and compiler diagnostics"
+input = """
+[1/4] Building CXX object src/ok.o
+[2/4] Building CXX object src/bad.o
+FAILED: src/bad.o
+/usr/bin/c++ -c ../src/bad.cc -o src/bad.o
+../src/bad.cc:10:5: error: use of undeclared identifier 'value'
+    value += 1;
+    ^
+ninja: build stopped: subcommand failed.
+"""
+expected = """FAILED: src/bad.o
+/usr/bin/c++ -c ../src/bad.cc -o src/bad.o
+../src/bad.cc:10:5: error: use of undeclared identifier 'value'
+    value += 1;
+    ^
+ninja: build stopped: subcommand failed."""
+
+[[tests.ninja]]
+name = "keeps compiler diagnostics and stopped summary"
+input = """
+[1/3] Building C object src/a.o
+[2/3] Building CXX object src/bad.o
+src/bad.cc:4:1: warning: control reaches end of non-void function
+ninja: build stopped: subcommand failed.
+"""
+expected = """src/bad.cc:4:1: warning: control reaches end of non-void function
+ninja: build stopped: subcommand failed."""
+
+[[tests.ninja]]
+name = "empty run collapses to ok message"
+input = """
+
+"""
+expected = "ninja: ok"


Koukyosyumei (approved):
## Summary

Clean, well-scoped filter addition that resolves #208. It follows the established `assets/filters/*.toml` conventions closely (compare `make.toml`, `gcc.toml`) and needs no Rust changes. I traced it against the engine in `src/filter_rules.rs` and mentally executed every golden case.

## What's good

- **All four golden tests pass.** I simulated `apply_filter`'s pipeline (strip_ansi → strip_lines_matching → truncate → max_lines → on_empty) against each case; all match after the `.trim()` comparison the harness uses (`src/filter_rules.rs:623`).
  - "successful build" → every `[n/m] Building|Linking` line stripped → empty → `on_empty` "ninja: ok". ✓
  - "keeps failed target" → progress lines stripped, `FAILED:` / compiler diagnostic / `ninja: build stopped` all kept. ✓
  - "keeps compiler diagnostics and stopped summary" → warning + stopped summary preserved. ✓
  - "empty run" → single blank line stripped → `on_empty`. ✓
- **Schema is valid.** Only known `FilterDef` fields are used, so `#[serde(deny_unknown_fields)]` (`src/filter_rules.rs:66`) is satisfied. Parses cleanly under `tomllib`.
- **Safe by construction.** No `match_output` rule, so this filter can never short-circuit to a fixed "ok" message and mask a real failure — the `can_hide_output` concern in `describe_file` (`filter_rules.rs:557`) doesn't apply. Failure output falls through untouched.
- **Routing is correct and conservative.** `^ninja\b` (via `command_string` basename normalization) matches `ninja`, `ninja -C build`, `/usr/bin/ninja all`, and the Fedora `ninja-build` binary, while correctly rejecting `ninja2 …` and `ninjas`.

## Non-blocking notes

1. **The progress-strip verb allowlist is generator-specific.** The second `strip_lines_matching` pattern only strips `[n/m]` lines whose description begins with a known verb (`Building|Linking|Compiling|…`). That covers CMake- and Meson-generated ninja well, but *raw* ninja or other generators emit descriptions like `[1/5] CC foo.o`, `AR libfoo.a`, `STAMP`, `GEN`, or a bare command line when a build edge has no `description`. Those progress lines won't be stripped, so token reduction is weaker there. This fails safe (signal is preserved, never hidden), so it's fine to ship — but since a `FAILED:` line never carries the `[n/m]` prefix, you could safely broaden to `^\[\d+/\d+\] ` and strip all progress regardless of generator, catching more noise with no added risk. Optional.

2. **No `filter_quality` coverage for a sizeable run.** The acceptance criteria mention "real token reduction on a sizeable run." The 4 inline golden tests satisfy the "2–3 golden tests" requirement, and `tests/filter_quality.rs` doesn't require per-filter entries, so this isn't blocking — but a large-ninja-log case there would guard the reduction property if you want it.

3. **`max_lines = 120`** is higher than siblings (`make`/`gcc` use 50). Reasonable for verbose parallel build output; just noting the inconsistency.

## Verification note

The author reported `cargo test builtin_golden_tests_pass` was blocked locally (no `cc` linker). I could not run the Rust suite either, but I reproduced the engine's line pipeline independently and confirmed all four goldens and the routing behavior pass. A CI run of `builtin_golden_tests_pass` should be green.

Overall: correct, safe, and consistent with the codebase. Approving; the notes above are optional polish.

PR #285, opened 2026-07-08
[codex] reject empty msg send and ask bodies

### LINKED ISSUE #278: Reject an empty/whitespace-only message body in `h5i msg send`/`ask`
## Reject an empty/whitespace-only message body in `h5i msg send` / `h5i msg ask`

`h5i msg send codex ""` (or a body that is only whitespace) currently records a pointless empty message into `refs/h5i/msg`. Fail-closed input validation is an explicit project value (see `CONTRIBUTING.md`: "Fail closed for … provenance ambiguity" and "Treat … messages … as untrusted"). A free-text message with no content should be refused with a clear error.

### Where
- Send path: `src/msg.rs`, `send_msg(...)` (around line 291) — it already validates `from`/`to` via `validate_name`, but never checks the `body`.
- CLI handler: `src/main.rs`, `MsgCommands::Send` (around line 5239), and the `ask` variant.

### Important nuance
Typed threaded replies **`ack` / `done` / `decline` legitimately allow an empty text** (the type carries the meaning). So the guard must live on the **free-text `send` / `ask` path**, not blanket inside `send_msg` where it would break the typed replies. Confirm which callers pass through the code you touch.

### What to do
1. In the `send` / `ask` path, reject a body that is empty after `.trim()` with a clear error, e.g. `message body must not be empty`.
2. Add a unit/integration test asserting the refusal for `""` and `"   "`.
3. Add/confirm a test that `ack` / `done` / `decline` with empty text still succeed (no regression).

### How to verify
```bash
cargo test --test msg_integration
h5i msg send codex ""        # should now error, not record an empty message
h5i msg send codex "hi"      # still works
```

### Acceptance criteria
- [ ] `send` / `ask` with an empty or whitespace-only body is refused with a clear message
- [ ] Typed replies (`ack` / `done` / `decline`) with empty text are unaffected
- [ ] Tests cover both the refusal and the still-allowed cases

leemeo3:
## Summary

- reject empty or whitespace-only free-text `h5i msg send` and `h5i msg ask` bodies with a clear error
- keep typed `ack` / `done` / `decline` replies with empty text allowed
- add integration coverage for the refusal path and typed reply regression case

Fixes #278

## Validation

- `cargo test --test msg_integration`
- `cargo clippy --bin h5i --test msg_integration -- -D warnings`
- `git diff --check`

AI assistance was used to prepare this change.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -619,6 +619,14 @@ fn send_reply(
     Ok(())
 }
 
+fn non_empty_free_text_body(parts: &[String]) -> anyhow::Result<String> {
+    let body = parts.join(" ");
+    if body.trim().is_empty() {
+        anyhow::bail!("message body must not be empty");
+    }
+    Ok(body)
+}
+
 /// Truncate a string to at most `max_chars` characters, appending `…` if cut.
 fn truncate(s: &str, max_chars: usize) -> String {
     let mut chars = s.chars();
@@ -5349,13 +5357,13 @@ fn main() -> anyhow::Result<()> {
                     branch,
                 }) => {
                     let me = msg::resolve_identity(&h5i_root, from.as_deref())?;
+                    let body = non_empty_free_text_body(&body)?;
                     let opts = msg::SendOpts {
                         tag,
                         branch,
                         ..Default::default()
                     };
-                    let sent =
-                        msg::send_msg(git, &h5i_root, &me, &to, &body.join(" "), opts)?;
+                    let sent = msg::send_msg(git, &h5i_root, &me, &to, &body, opts)?;
                     // Mirror to a confined recipient's per-env read-only inbox so a
                     // boxed team agent receives it (team id unknown here → match on
                     // the recipient agent). No-op if the recipient isn't boxed.
@@ -5370,19 +5378,13 @@ fn main() -> anyhow::Result<()> {
                     branch,
                 }) => {
                     let me = msg::resolve_identity(&h5i_root, from.as_deref())?;
+                    let body = non_empty_free_text_body(&body)?;
                     let opts = msg::SendOpts {
                         kind: Some("ASK".into()),
                         branch,
                         ..Default::default()
                     };
-                    report_sent(&msg::send_msg(
-                        git,
-                        &h5i_root,
-                        &me,
-                        &to,
-                        &body.join(" "),
-                        opts,
-                    )?);
+                    report_sent(&msg::send_msg(git, &h5i_root, &me, &to, &body, opts)?);
                 }
 
                 Some(MsgCommands::Review {
--- a/tests/msg_integration.rs
+++ b/tests/msg_integration.rs
@@ -331,6 +331,36 @@ fn reply_without_a_view_fails_clearly() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


Koukyosyumei (commented):
## Summary

This is a clean, well-scoped fix for #278. It rejects empty/whitespace-only free-text bodies on `h5i msg send` and `h5i msg ask` while intentionally preserving empty text for the typed `ack`/`done`/`decline` replies. The refactor into a single `non_empty_free_text_body` helper is tidy and the tests are solid. I traced the logic and it is correct; the note below is about scope consistency, not a blocker.

## What's good

- **Correct guard logic.** `non_empty_free_text_body` joins with `" "` then checks `trim().is_empty()`, so `[""]`, `["   "]`, and `["  ", "  "]` all bail, while `["hi"]` passes. It returns the *untrimmed* joined body, so legitimate messages with intentional surrounding spaces are stored exactly as before — no behavior regression. This complements clap's existing `required = true` (which only enforces ≥1 arg and happily accepts an empty-string arg), so the helper closes the real gap.
- **Guard runs before `send_msg`.** Nothing is written to `refs/h5i/msg` on rejection, which the test asserts precisely (`log.lines().count() == 1` after one valid send).
- **Refactor removes duplication cleanly.** Passing `&body` (now a `String`) into `send_msg` avoids the prior inline `body.join(" ")` and introduces no extra clones.
- **Good test coverage.** `send_and_ask_reject_empty_or_whitespace_body` checks both the non-zero exit and the exact stderr substring for send and ask, empty and whitespace. `ack_done_decline_allow_empty_reply_text` is a valuable regression test locking in that typed replies still accept empty bodies (`body == ""`, `reply_to`/`thread_id` present).

## Issues / notes (non-blocking)

### 1. Scope: other free-text senders remain unguarded (Low, consistency)
`reply`, `review`, `risk`, and `handoff` all take `#[arg(trailing_var_arg = true, required = true)] body: Vec<String>` and still call `msg::send_msg(..., &body.join(" "), ...)` / `send_reply(..., body.join(" "))` directly (src/main.rs ~5407, ~5433, ~5459, ~5472). They accept the same empty/whitespace bodies this PR now rejects for send/ask.

`reply` is the closest analog to send/ask — it is the *untyped* free-text reply (unlike the typed `ack`/`done`/`decline`, which deliberately allow empty text), so an empty `h5i msg reply <n> ""` producing a blank message is the same defect #278 describes. `review`/`risk`/`handoff` carry structured fields (`focus`/`risk`/`pr`/`priority`/`context`) so an empty prose body is more defensible there, but it is still inconsistent.

Suggestion: either extend `non_empty_free_text_body` to `reply` (and optionally review/risk/handoff), or add a one-line comment / PR note stating the scope is intentionally limited to `send`/`ask` per the issue. Reusing the helper for `reply` would be a two-line change.

### 2. Minor: helper naming (nit)
`non_empty_free_text_body` both validates and produces the joined body. The name reads fine; no change needed. Just flagging that if the guard is later reused for `reply`, keeping this single helper is the right call.

## Verdict rationale
Correct, minimal, well-tested, and matches the stated issue scope. I'm marking this a comment rather than an approval only to surface the `reply` gap for a decision (extend vs. document). Nothing here blocks merge.

PR #287, opened 2026-07-08
fix(cli): treat recall log --limit 0 as no limit

### LINKED ISSUE #283: Treat `h5i recall log --limit 0` as "no limit" instead of silently showing nothing
## Treat `h5i recall log --limit 0` as "no limit" (or reject it) instead of silently showing nothing

`h5i recall log --limit N` shows the N most recent commits (default 10). Passing `--limit 0` currently prints **nothing** and exits 0, with no explanation — the underlying `revwalk.take(limit)` simply takes zero items. That's a surprising, silent papercut: most CLIs treat `0` as "unlimited" (or reject it outright).

### Where
- Handler: `src/main.rs`, `Commands::Log { limit, ancestry }` (around line 6266).
- The value flows into `revwalk.take(limit)` in `src/repository.rs` (the log path).

### What to do (pick one, note it in the PR)
**Option A — `0` means unlimited (recommended, matches common convention):** when `limit == 0`, don't cap the revwalk (show the full history). Scope the change to the `recall log` path so other callers of the shared helper are unaffected — e.g. translate `0` to `usize::MAX` (or skip `.take`) in the `recall log` handler before it reaches the shared code.

**Option B — reject `0`:** return a clear error like `--limit must be at least 1`.

Either is a fine good-first-issue; A is the friendlier behavior. Keep the default (10) and all positive values behaving exactly as today.

### How to verify
```bash
h5i recall log --limit 0     # Option A: shows full history; Option B: clear error
h5i recall log --limit 3     # unchanged: 3 most recent
h5i recall log               # unchanged: default 10
```

### Acceptance criteria
- [ ] `--limit 0` no longer silently prints nothing
- [ ] Positive limits and the default are unchanged
- [ ] A test covers the `0` behavior you chose
- [ ] The chosen behavior is reflected in the flag's help text

ychampion:
## Summary

- Treat `h5i recall log --limit 0` as uncapped instead of rendering no commits.
- Update the flag help/manual text and regression coverage.

## Why

Fixes #283. This uses Option A from the issue: `0` means no limit. The CLI now translates that value before it reaches the shared log helper, so the default and positive limits keep their existing behavior.

## Validation

- `cargo test --test cli_integration recall_log_limit_caps_entries -- --nocapture` — passed
- `cargo test --test cli_integration` — passed
- `git diff --check` — passed
- `cargo test --lib -- --skip seccomp_notify::tests::live_socket_gate_denies_raw_allows_inet` — not completed locally; this host ran out of `/tmp` space during `mcp::tests::env_lifecycle_over_mcp` while copying plugin test fixtures

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -721,7 +721,7 @@ Show commit history with full AI provenance inline. Canonical form of the legacy
 
 | Option | Description |
 |--------|-------------|
-| `--limit <n>` | Number of commits to show (default: all) |
+| `--limit <n>` | Number of commits to show (default: 10; `0` = all) |
 | `--ancestry <file>:<line>` | Trace every commit that touched a specific line, annotated with its prompt |
 
 **Example — recent commits**
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -1067,7 +1067,7 @@ h5i hook codex finish [--summary &lt;text&gt;] [--quiet]
 <tbody>
 <tr>
 <td><code>--limit &lt;n&gt;</code></td>
-<td>Number of commits to show (default: all)</td>
+<td>Number of commits to show (default: 10; <code>0</code> = all)</td>
 </tr>
 <tr>
 <td><code>--ancestry &lt;file&gt;:&lt;line&gt;</code></td>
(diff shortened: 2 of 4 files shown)


Koukyosyumei (approved):
LGTM. Thank you so much!!

PR #289, opened 2026-07-08
feat(filters): add clang output filter

### LINKED ISSUE #282: Add an output filter for `clang` / `clang++`
## Add an output filter for `clang` / `clang++`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time.

There's already a filter for `gcc`/`g++` ([`assets/filters/gcc.toml`](https://github.com/h5i-dev/h5i/blob/main/assets/filters/gcc.toml)) but **none for `clang`/`clang++`**, which is just as common (default compiler on macOS, and widely used on Linux). This is a self-contained good-first-issue: **add a filter for `clang`**. No core-logic changes needed.

### Why
`clang`/`clang++` print include-chain context (`In file included from …`), caret/underline annotation lines, and `N warnings generated` / `N errors generated` footers; the signal is the actual error/warning diagnostics.

### What to keep / strip
- **Keep (the signal):** `error:` and `warning:` diagnostic lines.
- **Strip (the noise):** `In file included from …` chains, blank lines, and the `N warnings/errors generated` footers.

### Steps
1. Copy `assets/filters/gcc.toml` to `assets/filters/clang.toml` as a starting point — the noise patterns are nearly identical.
2. Set `match_command` to match clang invocations, e.g. `match_command = "^clang(\\+\\+)?\\b"`. Make it match typical invocations but not unrelated commands.
3. Adjust `strip_lines_matching` / `on_empty` as needed for clang's exact wording.
4. Add **2–3 inline golden tests** (`[[tests.clang]]`): a normal/clean run, an error run, and a warning run.

`description`, `match_command`, `strip_ansi`, `replace`, `match_output`, `strip_lines_matching`, `keep_lines_matching`, `truncate_lines_at`, `head_lines`, `tail_lines`, `max_lines`, `on_empty`.

### How to verify
```bash
```

### Acceptance criteria
- [ ] `assets/filters/clang.toml` added with a clear `description`
- [ ] `match_command` matches typical `clang`/`clang++` invocations and not unrelated ones
- [ ] At least 2 golden tests pass (normal + error/warning)
- [ ] Real token reduction on a sizeable run, with error/warning diagnostics preserved

zerone0x:
## Summary
Fixes #282

Adds a built-in declarative output filter for `clang` / `clang++`, mirroring the existing gcc/g++ behavior so compiler output keeps diagnostics while stripping include-chain/noise lines.

## Changes
- Adds `assets/filters/clang.toml` with inline golden cases.
- Adds clang/clang++ route coverage to filter rule tests.

## Verification
- `python3`/`tomllib` parse check for `assets/filters/clang.toml` ✅
- `cargo test filter_rules --quiet` attempted, but this runner lacks a C linker: `linker cc not found`.

### DIFF
--- a/assets/filters/clang.toml
+++ b/assets/filters/clang.toml
@@ -1,0 +1,54 @@
+[filters.clang]
+description = "Compact clang/clang++ compiler output — strip include notes, keep errors and warnings"
+match_command = "^clang(\\+\\+)?\\b"
+strip_ansi = true
+strip_lines_matching = [
+  "^\\s*$",
+  "^\\s+\\|\\s*$",
+  "^In file included from",
+  "^\\s+from\\s",
+  "^\\d+ warnings? generated",
+  "^\\d+ errors? generated",
+]
+max_lines = 50
+on_empty = "clang: ok"
+
+[[tests.clang]]
+name = "strips include chain, keeps errors and warnings"
+input = """
+In file included from /usr/include/stdio.h:42:
+                 from main.c:1:
+main.c:10:5: error: use of undeclared identifier 'foo'
+    foo();
+    ^
+main.c:15:12: warning: unused variable 'x' [-Wunused-variable]
+    int x = 42;
+        ^
+2 warnings generated.
+1 error generated.
+"""
+expected = "main.c:10:5: error: use of undeclared identifier 'foo'\n    foo();\n    ^\nmain.c:15:12: warning: unused variable 'x' [-Wunused-variable]\n    int x = 42;\n        ^"
+
+[[tests.clang]]
+name = "clean compilation"
+input = """
+"""
+expected = "clang: ok"
+
+[[tests.clang]]
+name = "linker error kept"
+input = """
+/usr/bin/ld: /tmp/main.o: undefined reference to 'missing_func'
+clang: error: linker command failed with exit code 1 (use -v to see invocation)
+"""
+expected = "/usr/bin/ld: /tmp/main.o: undefined reference to 'missing_func'\nclang: error: linker command failed with exit code 1 (use -v to see invocation)"
+
+[[tests.clang]]
+name = "clang++ command uses same rule"
+input = """
+hello.cc:3:10: fatal error: 'missing.hpp' file not found
+#include <missing.hpp>
+         ^~~~~~~~~~~~~
+1 error generated.
+"""
+expected = "hello.cc:3:10: fatal error: 'missing.hpp' file not found\n#include <missing.hpp>\n         ^~~~~~~~~~~~~"
--- a/src/filter_rules.rs
+++ b/src/filter_rules.rs
@@ -700,6 +700,8 @@ mod tests {
             (&["gradlew", "build"], "gradle"),
             (&["./gradlew", "build"], "gradle"),
             (&["gcc", "-O2", "main.c"], "gcc"),
+            (&["clang", "-Wall", "main.c"], "clang"),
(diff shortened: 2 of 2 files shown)


Koukyosyumei (approved):
## Summary

This PR adds a built-in `clang`/`clang++` output filter (`assets/filters/clang.toml`) plus routing/ambiguity test coverage in `src/filter_rules.rs`, closely mirroring the existing `gcc.toml`. The change is small, well-scoped, and correct. I traced every golden case and the routing regex by hand and empirically.

## What's good

- **Auto-registration works with no glue code.** Filters are embedded via `RustEmbed` from `assets/filters/` (`filter_rules.rs:35-38`) and any `*.toml` is picked up by `rule_paths()`, so adding the file is sufficient. No count/list constant needs updating (`reg.filters.len() >= 50` at `filter_rules.rs:651` still holds).
- **Golden cases are correct.** I executed all four `[[tests.clang]]` cases against `apply_filter` (`filter_rules.rs:248-338`):
  - *strips include chain* — `^In file included from` and `^\s+from\s` drop the two include lines; `^\d+ warnings? generated` / `^\d+ errors? generated` drop the trailing summary; error/warning/caret lines survive. Output matches `expected` exactly (including no trailing newline, since `.lines()` + `join("\n")` drops it).
  - *clean compilation* (`"""\n"""`) — TOML trims the leading newline to an empty body, so `on_empty = "clang: ok"` fires. Correct.
  - *linker error* and *clang++ fatal error* — nothing in the strip set matches the kept lines; the `1 error generated.` trailer is stripped. Both match.
- **The strip patterns are actually more accurate for clang than for gcc.** `N warnings/errors generated.` is genuinely clang output (gcc doesn't emit it), so copying the gcc rules lands on solid ground here.
- **Routing tests are meaningful.** The ambiguity loop (`filter_rules.rs:760`) confirms `clang main.c` / `clang++ main.cc` each match `<= 1` rule, and the route-map adds explicit `clang` / `clang++` assertions.

## Non-blocking notes

- **The `(\+\+)?` group in `match_command = "^clang(\\+\\+)?\\b"` is effectively dead.** For `clang++ …`, the `\b` after `++` can't match (both `+` and space are non-word chars), so the engine backtracks the optional group to empty and matches the `\b` at the `g`→`+` boundary right after `clang`. I confirmed `^clang(\+\+)?\b` matches `clang`, `clang++`, and `clang++ main.cc`. It works, but the group never actually changes the outcome. This is fine to leave; just noting it in case future readers assume it's load-bearing. (Interestingly, the analogous `gcc.toml` pattern `^g(cc|\+\+)\b` has the *opposite* latent issue — it does **not** match `g++` at all — but that's pre-existing and out of scope here.)
- **Broad `clang*` matching.** Because of that `\b`, `clang-tidy`, `clang-format`, `clang-cl`, and `clang-15` also route to this compiler filter. The strip rules are conservative so it's mostly harmless, but `max_lines = 50` could truncate a long `clang-tidy` run. This matches the existing breadth of `gcc.toml` (`gcc-ar`, etc.), so it's consistent with project convention rather than a new problem — worth a mental note only.

## Verdict

Correct, well-tested, and consistent with the sibling `gcc.toml`. The PR's own note that `cargo test filter_rules` couldn't run (no C linker in the runner) doesn't matter — the golden/routing tests here are pure-Rust string logic with no compiler invocation, so CI will exercise them fully. Approving.

PR #290, opened 2026-07-08
test: add filter edge-case goldens

### LINKED ISSUE #211: Add error/edge-case golden tests to thinly-tested filters
## Add a third golden test (error / alternate format) to filters that have only two

Most filters in [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters) ship with just two golden tests (typically a success case and one variant). Adding an **error/failure** case and, where relevant, an **alternate-format** case (e.g. Windows vs. Unix output) hardens them against regressions.

### What to do
Pick a batch of filters (e.g. `npm`, `pip`, `go`, `gradle`, `mypy`, `flake8`, `jest`, `yarn`) and add an `[[tests.<name>]]` case that feeds an **error/failure** raw output and asserts the filter keeps the error/signal while dropping the noise. Use [`ping.toml`](https://github.com/h5i-dev/h5i/blob/main/assets/filters/ping.toml) (which has a success, an alternate-format, and an error case) as the model.

### Verify
```bash
cargo test builtin_golden_tests_pass
```

### Acceptance criteria
- [ ] At least 6 filters gain a meaningful error/edge-case golden test
- [ ] Each new test demonstrably keeps the error signal and drops noise
- [ ] `cargo test builtin_golden_tests_pass` passes

zerone0x:
## Summary
Fixes #211

Adds meaningful error/edge-case golden coverage for six thinly-tested built-in filters.

## Changes
- Add npm dependency-resolution failure golden coverage
- Add pip wheel-build failure golden coverage
- Add Go compile-error golden coverage
- Add flake8 syntax-error/blank-line golden coverage
- Add Jest setup-failure golden coverage
- Add Yarn engine-mismatch golden coverage

## Verification
- `cargo test builtin_golden_tests_pass` (blocked locally: rustc could not find linker `cc` in this environment)
- `python3` TOML parse + local rule-emulator check for the six touched filters: passed (18 cases checked; all filter TOML files parse)

### DIFF
--- a/assets/filters/flake8.toml
+++ b/assets/filters/flake8.toml
@@ -24,3 +24,17 @@ src/a.py:10:80: E501 line too long (88 > 79 characters)
 name = "clean run collapses to message"
 input = ""
 expected = "flake8: clean"
+
+[[tests.flake8]]
+name = "syntax error keeps filename and code while dropping blanks"
+input = """
+
+src/bad.py:3:7: E999 SyntaxError: invalid syntax
+
+src/bad.py:4:1: F821 undefined name 'result'
+
+"""
+expected = """
+src/bad.py:3:7: E999 SyntaxError: invalid syntax
+src/bad.py:4:1: F821 undefined name 'result'
+"""
--- a/assets/filters/go.toml
+++ b/assets/filters/go.toml
@@ -45,3 +45,21 @@ PASS
 ok  \texample/m\t0.012s
 """
 expected = "go: ok"
+
+[[tests.go]]
+name = "compile error keeps package and source diagnostics"
+input = """
+go: downloading golang.org/x/text v0.14.0
+# example.com/app/cmd
+cmd/main.go:12:9: undefined: missingSymbol
+cmd/main.go:13:2: declared and not used: value
+FAIL\texample.com/app/cmd [build failed]
+FAIL
+"""
+expected = """
+# example.com/app/cmd
+cmd/main.go:12:9: undefined: missingSymbol
+cmd/main.go:13:2: declared and not used: value
+FAIL\texample.com/app/cmd [build failed]
+FAIL
+"""
--- a/assets/filters/jest.toml
+++ b/assets/filters/jest.toml
@@ -42,3 +42,24 @@ PASS src/b.test.ts
 Tests:       13 passed, 13 total
 """
 expected = "Tests:       13 passed, 13 total"
+
+[[tests.jest]]
+name = "setup failure keeps suite error and summary"
+input = """
+PASS src/ok.test.ts
+FAIL src/config.test.ts
+  ● Test suite failed to run
+    Cannot find module './missing-config' from 'src/config.test.ts'
+    > 1 | import './missing-config'
+A worker process has failed to exit gracefully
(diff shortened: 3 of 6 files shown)


Koukyosyumei (approved):
## Summary

Test-only PR (`+106/-0`, 6 files) adding one failure/edge-case golden per built-in filter (flake8, go, jest, npm, pip, yarn), closing #211. No engine or production code changes.

## Verification performed

I hand-executed every new case through `apply_filter` in `src/filter_rules.rs` (stages: `strip_ansi` → `strip_lines_matching` → `max_lines` → `on_empty`, compared with `.trim()` equality per `run_golden_tests`, `src/filter_rules.rs:619-633`) and then confirmed it empirically: applied the six files and ran `filter_rules::tests::builtin_golden_tests_pass` — **all pass** (the linker issue the author hit locally does not reproduce here). All six files also parse cleanly under a strict TOML parser (3 cases each).

Spot-checks worth noting:
- **flake8**: interleaved blank lines correctly collapse via `^\s*$`; leading/trailing blanks handled by the trim-compare. Correct.
- **go**: `^go: downloading ` strips the fetch line while `# pkg` / `cmd/main.go:…` diagnostics and both `FAIL` lines survive; the literal `\t` escapes match on both sides. Correct.
- **jest**: `A worker process has failed` is an *unanchored* strip pattern and correctly drops the "…failed to exit gracefully" line while keeping the `● Test suite failed to run` block and both summary lines. Good coverage of a substring rule.
- **npm**: `^npm WARN` + `^npm notice` stripped, all four `npm ERR!` lines kept. Correct.
- **yarn**: `^warning ` drops the invalid-engine warning while the `error package-b@…` and `error Found incompatible module.` lines survive; embedded double-quotes are valid inside the multiline basic string. Correct.

## What's good

- Cases are genuinely meaningful, not filler: each exercises the specific strip rules under a real failure shape rather than re-testing the happy path, which is exactly what #211 asked for.
- Style matches the existing `[[tests.<filter>]]` convention exactly (name/input/expected), so they slot into the existing harness with zero engine changes.
- The **pip** case is a nice anchor-boundary probe: `^\s*Building wheel` strips the `Building wheel for … ... error` progress line but correctly does *not* strip the `  × Building wheel …` summary line (the `×` breaks the anchored match), so the failure detail is preserved. This pins down a subtle rule interaction.

## Non-blocking observations

- In the **pip** case, the input line `  Building wheel for broken-package (pyproject.toml) ... error` (the one carrying the per-wheel `... error` marker) is itself stripped by `^\s*Building wheel`. The failure is still surfaced via the retained `error: subprocess-exited-with-error` and `ERROR: Failed building wheel …` lines, so the golden is correct and documents real behavior — just flagging that the test happens to encode a slightly lossy strip, which is fine but worth being aware of if the pip rule is ever tuned.
- All six additions are "keep-the-error" cases (the gap the issue targeted). Complementary edge cases like `match_output` short-circuits, `max_lines` truncation, or `truncate_lines_at` remain uncovered for these filters — reasonable to leave out of this PR's scope, but a candidate for a follow-up if broader coverage is desired.

No correctness, convention, performance, or security concerns. Recommend merge.

PR #292, opened 2026-07-08
fix(serve): friendly error when the dashboard port is already in use

### LINKED ISSUE #277: Give `h5i serve` a friendly error when the port is already in use
## Give `h5i serve` a friendly error when the port is already in use

When the dashboard port is taken, `h5i serve` fails with a raw, unhelpful OS error:

```
Error: Address already in use (os error 98)
```

There's no hint that `--port` exists. This is a common, self-inflicted papercut for a UI command, and an easy, well-scoped fix.

### Where
`src/server.rs`, at the listener bind (around line 3262):

```rust
let listener = TcpListener::bind(&addr).await?;
```

### What to do
Inspect the bind error and, when the kind is `std::io::ErrorKind::AddrInUse`, return/print a clear message such as:

```
port 8080 is already in use — pick another with `h5i serve --port <N>`
```

Keep the existing `?` fallthrough for all other error kinds (permission denied, invalid address, etc.) so only the "in use" case gets the special message.

### Sketch
```rust
let listener = match TcpListener::bind(&addr).await {
    Ok(l) => l,
    Err(e) if e.kind() == std::io::ErrorKind::AddrInUse => {
        // return a friendly h5i error mentioning --port and the actual port
    }
    Err(e) => return Err(e.into()),
};
```

### How to verify
1. Start something on the port (or a first `h5i serve`).
2. Run a second `h5i serve` on the same port and confirm the friendly message appears with the right port number and the `--port` hint.

### Acceptance criteria
- [ ] `AddrInUse` produces a clear message naming the port and suggesting `--port`
- [ ] Other bind errors are unchanged (still surfaced)
- [ ] No change to the successful path

AnayGarodia:
## What

`h5i serve` now gives a clear, actionable error when the dashboard port is already in use instead of a raw OS error.

Before:
```
Error: Address already in use (os error 98)
```
After:
```
Error: port 7150 is already in use — pick another with `h5i serve --port <N>`
```

## Why

Fixes #277. For a UI command this is a common, self-inflicted papercut, and the raw error gives no hint that `--port` exists.

## How

In `server::serve`, match on the `TcpListener::bind` result and special-case only `std::io::ErrorKind::AddrInUse`, returning an `anyhow` error that names the actual port and points at `h5i serve --port <N>`. Every other bind error (permission denied, invalid address, …) keeps the existing `?` fallthrough, and the successful path is unchanged.

## Testing

Added a unit test (`server::tests::serve_reports_friendly_error_when_port_in_use`) that occupies an OS-assigned port, calls `serve` on it, and asserts the error names the port, says "already in use", and mentions `--port`.

```
cargo test --lib serve_reports_friendly   # passes
cargo clippy --all-targets --all-features -- -D warnings   # clean
```

## Compatibility

No CLI flags, commands, stored refs, or output on the success path change — only the failure message for the in-use case. No docs/manual regeneration needed (the clap command tree is untouched).

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -3259,7 +3259,15 @@ pub async fn serve(repo_path: PathBuf, port: u16) -> anyhow::Result<()> {
     let app = build_router(state);
 
     let addr = format!("127.0.0.1:{}", port);
-    let listener = TcpListener::bind(&addr).await?;
+    let listener = match TcpListener::bind(&addr).await {
+        Ok(listener) => listener,
+        Err(e) if e.kind() == std::io::ErrorKind::AddrInUse => {
+            anyhow::bail!(
+                "port {port} is already in use — pick another with `h5i serve --port <N>`"
+            );
+        }
+        Err(e) => return Err(e.into()),
+    };
     println!("  h5i UI →  http://{}", addr);
     axum::serve(listener, app).await?;
     Ok(())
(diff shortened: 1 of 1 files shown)


Koukyosyumei (approved):
## Summary

Small, well-scoped UX fix: `serve` now converts the `AddrInUse` bind failure into an actionable message that names the port and points at `--port`, while every other bind error keeps the existing behavior. Clear PR description, and it ships a real regression test. Nice.

## What's good

- **Minimal and surgical.** The match on `TcpListener::bind` special-cases only `ErrorKind::AddrInUse`; the fallthrough `Err(e) => return Err(e.into())` is behaviorally identical to the original `?` (since `std::io::Error: Into<anyhow::Error>`), so permission-denied/invalid-addr errors are unchanged. Success path is untouched (server.rs:3257).
- **The test is correct and non-flaky.** `serve_reports_friendly_error_when_port_in_use` binds `127.0.0.1:0`, keeps the listener alive, and calls `serve` on that port. On Linux a second `bind` to an actively-listening `(addr, port)` returns `EADDRINUSE` regardless of `SO_REUSEADDR` (that flag only covers TIME_WAIT / wildcard cases), so the `AddrInUse` arm is hit deterministically. Bind fails before `axum::serve`, so the test returns promptly with no dangling server.
- **Consistent with conventions.** `use super::*` correctly resurfaces `TcpListener` (server.rs:13) and `serve`; the inline `{port}` capture in `anyhow::bail!` is valid. The em-dash in the message matches existing CLI error copy elsewhere in the tree (e.g. `main.rs`'s `"$HOME is not set — use --scope project"`), so it's in-house style, not a nit. This is also the first `#[cfg(test)] mod tests` in `server.rs`, so no module conflict.
- **Feature gating holds.** `server` is behind the default-on `web` feature (lib.rs:45), so the new test compiles and runs under the default `cargo test --verbose` that CI uses.

## Optional / non-blocking

- The test asserts on the three salient substrings (`already in use`, the port, `--port`) rather than the exact string — good, that keeps it resilient to copy tweaks. No change needed; just noting it's the right call.
- Consider (later, not here) a tiny doc-comment on `serve` noting the friendly-error contract so a future refactor doesn't silently drop it. Purely optional.

No correctness, security, or performance concerns. LGTM.

PR #294, opened 2026-07-08
Add --json output to env inspect

### LINKED ISSUE #281: Add `--json` output to `h5i env inspect`
## Add `--json` output to `h5i env inspect`

`h5i env inspect <name> --capture <id>` renders one evidence capture in a human-readable view, but has no machine-readable option. Sibling `env` commands (`status`, `list`, `compare`) expose `--json`; a `--json` here would let tooling consume a single capture (command, exit code, action, redactions, wall/cpu/rss accounting) without scraping the rendered text.

### Where
`src/main.rs`, the `EnvCommands::Inspect` variant (in the `EnvCommands` enum, around lines 1957+) and its handler.

### What to do
1. Add a `#[arg(long)] json: bool` to the `Inspect` variant.
2. When set, serialize the capture record with `serde_json` instead of rendering the pretty view.
3. Reuse the capture/object types already used by the human renderer so the JSON stays faithful to what's stored; follow `env status --json` for style.

Keep the redaction guarantees intact — the JSON must contain the same secret-redacted fields the human view shows, never raw secrets.

### How to verify
```bash
h5i env create demo && h5i env run demo -- echo hi
id=$(h5i recall objects --env demo --json | jq -r '.[0].id')   # grab a capture id
h5i env inspect demo --capture "$id" --json | jq .
h5i env inspect demo --capture "$id"        # human view unchanged
```

### Acceptance criteria
- [ ] `h5i env inspect <name> --capture <id> --json` emits valid JSON of the capture
- [ ] Secret redaction is preserved in the JSON output
- [ ] The default human view is unchanged

ychampion:
## Summary

- Add `--json` to `h5i env inspect`.
- Reuse the stored capture manifest and existing ownership check for the JSON path.
- Document the flag and refresh the generated manuals.

## Why

Closes #281. `env inspect` had a human view only, so tooling had to scrape rendered text to consume a single capture.

## Validation

- `cargo test --test env_integration inspect_json_renders_the_capture_manifest -- --nocapture` - passed
- `cargo test --test env_integration inspect_ -- --nocapture` - passed
- `scripts/gen_man.sh` - passed
- `python3 scripts/gen_manual.py` - passed
- `git diff --check -- . ':(exclude)man/man1/h5i.1'` - passed

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -2892,7 +2892,7 @@ another reviews and applies). See `docs/environments-design.md` and the live
 | `h5i env ports <name> [--json]` | The per-env **injected** port map: each running service with a declared port and the free host port h5i allocated and injected as `PORT` / `H5I_ENV_PORT_<NAME>`. v1 is injection only — there is **no host→box forwarder**, so a port is reachable only if the service binds the injected value (the URL is shown as conditional, not a guarantee). |
 | `h5i env log <name>` | The event log (`created`/`exec`/`service`/`proposed`/`applied`/`aborted`/`gc`/`violation`/`secret`). |
 | `h5i env diff <name> [--stat]` | Diff the env's work against its pinned base. |
-| `h5i env inspect <name> --capture <id>` | Render one evidence capture (structured findings, exit code, policy digest, redactions). |
+| `h5i env inspect <name> --capture <id> [--json]` | Render one evidence capture (structured findings, exit code, policy digest, redactions). `--json` emits the stored capture manifest for tooling instead of the human view. |
 | `h5i env compare <names…> [--json]` | The "arena": rank N envs side by side (changes + latest run results). Best on envs sharing one base. |
 | `h5i env rebase <name>` | Re-pin the base onto the parent branch's advanced tip (3-way; refuses on conflict). |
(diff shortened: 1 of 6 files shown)


Koukyosyumei (approved):
**Approve.** Clean, well-scoped addition of `--json` to `h5i env inspect`.

Strengths:
- The refactor is the right shape: `inspect_manifest` isolates the manifest resolution + env-ownership check, and `inspect` delegates to it. The human and JSON paths therefore share one code path, so they can't drift and the cross-env refusal is enforced identically for both. `objects::Manifest` already derives `Serialize`, so `to_string_pretty` is sound.
- Test coverage is on point: `inspect_json_renders_the_capture_manifest` asserts the concrete fields (`id`, `env_id`, `exit_code`, `policy_digest`, `raw_oid` prefix, `summary`) and, importantly, that a capture id from another env is still refused via `--json`. That last check is what keeps this from becoming an evidence-read escape hatch.
- All three doc surfaces (MANUAL.md, HTML manual, man page) are regenerated consistently, and clippy `--all-targets --all-features -D warnings` is green.

No blocking concerns. Nice, minimal change.

PR #295, opened 2026-07-09
Add --json output to env diff

### LINKED ISSUE #280: Add `--json` output to `h5i env diff`
## Add `--json` output to `h5i env diff`

`h5i env diff` currently exposes only `--stat` (a human diffstat) or the full patch text. Its sibling commands `env status`, `env list`, and `env compare` all support `--json`. A reviewer or tool comparing candidate environments gets JSON from those commands but has to scrape patch/`--stat` text from `env diff`.

Adding a structured `--json` diffstat closes that gap.

### Where
`src/main.rs`, the `EnvCommands::Diff` variant (in the `EnvCommands` enum, around lines 1957+) and its handler.

### What to do
1. Add a `#[arg(long)] json: bool` to the `Diff` variant (alongside the existing `--stat`).
2. When `json` is set, emit a structured diffstat instead of text, reusing the data already computed for `--stat`. A simple, useful shape:
   ```json
   { "files": [ { "path": "src/foo.rs", "insertions": 12, "deletions": 3 } ],
     "insertions": 12, "deletions": 3 }
   ```
3. Use `env status --json` / `env compare --json` as the reference for serialization style.

Scope this issue to `env diff` only. (`env inspect --json` is tracked separately.)

### How to verify
```bash
h5i env create demo && h5i env run demo -- sh -c 'echo hi > f.txt'
h5i env diff demo --json | jq .    # structured diffstat
h5i env diff demo --stat           # unchanged human view
```

### Acceptance criteria
- [ ] `h5i env diff <name> --json` emits a valid structured diffstat
- [ ] `--stat` and the default patch output are unchanged
- [ ] The JSON shape is documented in the flag help and/or `MANUAL.md`

ychampion:
## Summary

- Add `--json` to `h5i env diff` with per-file and total diffstat counts.
- Cover both local worktree and pulled remote-env diff paths.
- Update the manual, HTML manual, and man page.

## Why

Closes #280. `env diff` had patch and human diffstat output, but no structured form for tooling.

## Validation

- `cargo test --test env_integration full_lifecycle_create_run_propose_apply -- --nocapture` - passed
- `cargo test --test env_integration env_travels_to_another_clone_for_review_and_apply -- --nocapture` - passed
- `scripts/gen_man.sh` - passed
- `python3 scripts/gen_manual.py` - passed
- `git diff --check -- . ':(exclude)man/man1/h5i.1'` - passed


Koukyosyumei (approved):
**Approve.** Solid structured mirror of `h5i env diff` via `--json`.

Strengths:
- `diffstat_report` faithfully mirrors the existing `diff()` for both branches: identical `base_tree` resolution, the same `DiffOptions` (untracked included / recursed / content shown) on the local-worktree path, and the same tip-tree resolution plus the "env code branch not present locally — pull it first" error on the remote/no-worktree path. Keeping the two in lockstep means the JSON stat and the human diff describe exactly the same change set.
- Graceful edge handling: `Patch::from_diff(...).map(...).transpose()?.unwrap_or((0, 0, 0))` means binary or unparseable files fall back to zero line counts rather than erroring, and delta path selection correctly prefers `old_file()` for deletions and `new_file()` otherwise.
- Tests cover both the local lifecycle path (per-file + total insertion/deletion counts on `lib.py`) and the cross-clone remote path in `env_travels_to_another_clone_for_review_and_apply`. Docs regenerated across all three surfaces; clippy `-D warnings` green.

Minor, non-blocking: `diffstat_report` calls `Patch::from_diff` once per delta to get per-file line stats, which re-parses each file's hunks on top of the aggregate `diff.stats()`. Fine at env-diff scale; just noting it's O(patch) per file if a future caller runs it on very large diffs. No change needed for merge.

PR #296, opened 2026-07-09
Fix captured test evidence in team submissions

### LINKED ISSUE #224: team: artifact show --tests is empty — submitter's test captures not linked to the submission
## Problem

`h5i team artifact show <id> --tests` always reports `(no captured test evidence)`, even when the submitting agent did run its test suite. So a reviewer who was granted the `tests` artifact (`granted: diff,summary,tests`) cannot actually see any test results — which partly defeats that grant in the peer-review flow.

This recurs across every team experiment run (T2/T3/T4).

## Evidence (T4 logs, `log3/codex.txt`)

The submitter ran its tests under capture:

```
• Ran h5i capture run -- python3 -m pytest
  └ ... 5 passed in 0.01s
```

But when the reviewer inspects the submission:

```
• Ran h5i capture run -- h5i team artifact show sub-rohan-r1-99b2c23144d5 --tests
  └ submission sub-rohan-r1-99b2c23144d5 · rohan · round 1 · 1 file(s) +86 -0
      (no captured test evidence)
```

## Likely cause

`TeamArtifact.capture_ids` (`src/team.rs`) is populated from the env manifest's captures (`m.captures.clone()`) at submit time, but the agent's in-box test-run captures appear not to be associated with the submission by then:

- A boxed `h5i team agent submit` stages a `TeamSubmitSpool`; the host builds the artifact during `ingest_team_outbound` / `ingest_shell_spool`.
- The agent's `python3 -m pytest` capture is staged in the env spool and ingested into the env's captures, but the submission artifact's `capture_ids` ends up empty.

Need to confirm ordering: are the inbox captures ingested into `m.captures` **before** the submission artifact is built host-side? If not, the artifact misses them. Also consider scoping which captures count as "test evidence" (e.g. the test-kind captures, or those tagged `--tests`).

## Where to look

- `src/team.rs` — `submit()` building `TeamArtifact { capture_ids: m.captures.clone(), ... }`; `TeamArtifact.capture_ids`.
- `src/env.rs` — `ingest_team_outbound` / `ingest_shell_spool` (the host-side ingest order: captures vs. submission).
- `src/main.rs` — `TeamArtifactCommands::Show` `--tests` rendering of `art.capture_ids`.

## Acceptance

A reviewer running `h5i team artifact show <id> --tests` sees the submitter's captured test run(s) (capture ids + structured status/counts), so the `tests` grant is meaningful. Add a test covering: submit after a captured test run → the submission's `capture_ids` includes it → `artifact show --tests` renders it.

_Surfaced during T4 team-run log analysis; deferred from the in-box review-friction fixes (artifact show, staged-capture rehydrate, read-through capture passthrough)._

ychampion:
## Summary

- Persist the env manifest after shell/capture spool ingest before draining team-submit records.
- Add a regression for a captured test run and `team agent submit` staged in the same spool.

## Why

Closes #224. The team submit path reloads the env manifest while building `TeamArtifact.capture_ids`; without the intermediate save, same-session captured test evidence is missing from the submission artifact.

## Validation

- `cargo test shell_ingest_links_captured_tests_to_team_submission --lib -- --nocapture` - failed before the fix, passed after
- `cargo test write_team_submit_spool_records_scoped_request --lib -- --nocapture` - passed
- `cargo test auto_peer_review_ingests_staged_submission_before_freezing --lib -- --nocapture` - passed
- `cargo test sync_outbound_ingests_staged_submission_live --lib -- --nocapture` - passed
- `cargo test submit_auto_snapshots_dirty_worktree --lib -- --nocapture` - passed
- `git diff --check` - passed

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -3882,6 +3882,11 @@ fn ingest_shell_spool(
         let _ = std::fs::remove_file(&path);
     }
 
+    // Captures ingested above only live in this mutable manifest until the
+    // caller's final status write. Team submission ingest reloads the env
+    // manifest, so persist first or the submission misses same-spool evidence.
+    save_manifest(h5i_root, m)?;
+
     // Drain the in-box team outbound spool (submissions + peer reviews); the
     // same path runs on demand via `h5i team sync` (see `ingest_team_outbound`).
     count += ingest_team_outbound(repo, h5i_root, m)?;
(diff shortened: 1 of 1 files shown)


Koukyosyumei (approved):
**Approve.** Correct, minimally-scoped fix for a real same-session evidence-loss bug (#224).

Root cause is precisely diagnosed: `ingest_shell_spool` appends each freshly-ingested capture into the in-memory `m.captures`, but those only reach disk via the caller's *final* status write. The team-submit drain that runs at the tail of the same function (`ingest_team_outbound` → `team::submit`) reloads the manifest from disk (`env::find`) and builds `capture_ids: m.captures.clone()` from that reloaded copy — so before this change it saw a stale manifest and dropped the same-spool test evidence. Inserting `save_manifest(h5i_root, m)?` immediately before the drain is the right place: the drain lives inside `ingest_shell_spool`, so the persist has to happen here, not in the caller.

Why it's safe:
- `ingest_team_outbound` takes `m` by `&EnvManifest` (immutable), so it cannot further mutate the manifest — the intermediate save and the caller's later final save stay consistent (captures are only appended, never removed). The extra write is negligible I/O.
- Fail-closed: if the persist errors, the whole ingest fails before submitting, so a submission never goes out with evidence that couldn't be durably recorded.
- The explanatory comment captures the non-obvious "reload sees only what's on disk" invariant for the next reader.

The regression test is a genuine end-to-end reproduction: it stages a pytest capture spool *and* a `team-submit` spool together, runs the ingest, and asserts both that the submission carries the one capture id and that the persisted manifest's captures equal the submission's — so it locks in the ordering, not just the count. PR notes confirm it fails before / passes after, and clippy `-D warnings` is green.

No blocking concerns.

PR #306, opened 2026-07-09
feat(env): add JSON output for env log

### LINKED ISSUE #279: Add `--json` to `h5i env log` for parity with the env family
## Add `--json` to `h5i env log` for parity with the rest of the `env` family

Most of the `h5i env` subcommands expose a `--json` flag for machine-readable output — `env list`, `env status`, `env compare`, `env capabilities`, `env ports` all do. **`env log` does not**, so tooling that parses the env family as JSON has to fall back to scraping its human-formatted text.

### Where
- Definition: `src/main.rs`, the `EnvCommands::Log` variant — currently `Log { name: String }` (around line 1974), with no `json` field.
- Handler: `src/main.rs`, `EnvCommands::Log` match arm (around line 11392).

`env log` already reads structured event records from `refs/h5i/env`, so emitting them as JSON is a natural addition.

### What to do
1. Add a `#[arg(long)] json: bool` to the `Log` variant (mirror how `Ports { name, json }` is declared right nearby).
2. In the handler, when `json` is set, serialize the event records with `serde_json` instead of printing the pretty view.
3. Follow the existing `env status --json` handler as the reference for shape and error handling.

### How to verify
```bash
h5i env create demo && h5i env run demo -- true
h5i env log demo --json | jq .     # should be valid JSON of the event records
h5i env log demo                   # human view unchanged
```

### Acceptance criteria
- [ ] `h5i env log <name> --json` emits valid JSON of the event records
- [ ] The default (non-`--json`) output is unchanged
- [ ] Help text for the flag matches the style of the other `env … --json` flags

ychampion:
## Summary

- add `--json` to `h5i env log`
- serialize the environment's stored event records while preserving the existing text view
- cover event scoping, lifecycle records, capture IDs, and unchanged human output

## Why

The rest of the `env` command family exposes machine-readable output, but `env log` still requires callers to scrape its text view. The event records are already structured, so this exposes them directly without introducing a second output shape.

Fixes #279

## Validation

- `cargo test --test env_integration log_json_emits_env_event_records -- --nocapture` — passed
- `scripts/gen_man.sh` — passed, 0 groff warnings
- `python3 scripts/gen_manual.py` — passed


Koukyosyumei (approved):
Reviewed the diff and verified locally: applied the patch on top of main, ran the new integration test (`cargo test --test env_integration log_json_emits_env_event_records`) — passes.

**What it does:** adds `--json` to `h5i env log`, serializing the env-scoped `EnvEvent` records directly instead of the text view.

**Correctness / conventions:**
- `EnvEvent` already derives `Serialize` with `skip_serializing_if` on the optional `detail`/`capture` fields, so the JSON shape is clean and stable.
- `serde_json::to_string_pretty` matches every other `--json` in the env family (`status`, `diff`, `list`, `ports`, …) — no second output shape introduced.
- No new information exposure: `detail` is secret-redacted at write time, and the only fields JSON adds over the text view are `env_id`/`agent`, which are not sensitive.
- Test coverage is good: asserts env scoping, lifecycle (`created`), capture evidence on `exec`, and that the human view is unchanged.
- Docs regenerated consistently (MANUAL.md, HTML manual, man page). The man-page trailing-whitespace warning on the SYNOPSIS line is a pre-existing generator artifact, not introduced here.

LGTM. ✅

PR #307, opened 2026-07-10
docs: document the top-level `h5i doctor` command

### LINKED ISSUE #275: Document the top-level `h5i doctor` command in MANUAL.md
## Document the top-level `h5i doctor` command in `MANUAL.md`

`h5i doctor` validates and repairs h5i's sidecar storage and refs, but it is the **only top-level command missing from `MANUAL.md`** — a user reading the full command reference will never discover it. (`MANUAL.md` documents `h5i env doctor`, but not the top-level `h5i doctor`.)

### What the command does
Verified via `h5i doctor --help`:

```
Validate and repair h5i sidecar storage and refs

Options:
      --repair        Create missing sidecar directories and schema metadata
      --export <DIR>  Export a recovery copy of .git/.h5i plus a refs manifest into this directory
      --json          Output raw JSON instead of the pretty report
```

Implementation lives in `src/main.rs` (the top-level `Doctor { .. }` variant, around line 1022, and its handler).

### Steps
1. Add a `h5i doctor` section to `MANUAL.md` in the same style as neighboring top-level command sections.
2. Describe each flag: `--repair` (creates missing sidecar dirs/schema), `--export <DIR>` (writes a recovery copy of `.git/.h5i` + a refs manifest), `--json` (raw report for tooling).
3. If `MANUAL.md` is mirrored to HTML under `docs/manual/`, regenerate/update the mirror too (check how other commands are kept in sync).

### How to verify
```bash
grep -n "h5i doctor" MANUAL.md          # should now show the new section
h5i doctor --help                       # confirm your flag descriptions match
```

### Acceptance criteria
- [ ] A `h5i doctor` section exists in `MANUAL.md`
- [ ] All three flags (`--repair`, `--export`, `--json`) are documented accurately
- [ ] Style matches the surrounding command reference entries

AnayGarodia:
Closes #275.

## What

Adds a `h5i doctor` section to `MANUAL.md` (Core commands group), plus a TOC entry and the regenerated `docs/manual/index.html` mirror.

## Why

`h5i doctor` was the only top-level command missing from the manual — the reference documents `h5i env doctor` but not the top-level one, so a reader had no way to discover it.

## How

New section covering `--repair`, `--export <DIR>`, and `--json`, with flag descriptions taken from `h5i doctor --help`, and a cross-reference disambiguating it from the per-env `h5i env doctor`. Regenerated the HTML mirror via `scripts/gen_manual.py` (additive, no churn).

AI-assisted; verified locally that the cross-reference anchor resolves and the mirror renders the new section.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -76,6 +76,7 @@ emits `--json` is safe to consume from a script or another agent.)
 - [Core commands](#core-commands)
   - [h5i init](#h5i-init)
   - [h5i resolve](#h5i-resolve)
+  - [h5i doctor](#h5i-doctor)
 - [h5i capture](#h5i-capture)
   - [h5i capture commit](#h5i-capture-commit)
   - [h5i capture memory](#h5i-capture-memory)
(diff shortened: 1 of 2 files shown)


Koukyosyumei (approved):
Reviewed and verified against the implementation:

- **Flags match the CLI**: `--repair` / `--export <DIR>` / `--json` descriptions line up with the clap definitions in `main.rs` (Commands::Doctor), and `--repair`'s "only fills in what is absent" matches `storage::doctor` (creates missing dirs/metadata, never touches existing data).
- **Exit-code claim is accurate**: the command exits non-zero (2) when `report.ok` is false, so gating scripts on it is safe as documented.
- **Cross-reference anchor is right**: `h5i env doctor` has no dedicated section (it's a table row in the `h5i env` section), so linking `#h5i-env` is the correct target.
- **HTML mirror is a faithful regen**: I applied the patch and re-ran `scripts/gen_manual.py` locally — the output is byte-identical to the committed `docs/manual/index.html` (27 insertions, no churn).

Nice, accurate docs addition. LGTM.

PR #308, opened 2026-07-10
test(filters): golden-test the on_empty path for the remaining filters

### LINKED ISSUE #210: Add clean/empty-run golden tests to filters that define on_empty
## Golden-test the `on_empty` path for every filter that sets it

Several filters declare an `on_empty = "…"` message (the compact line emitted when a command produces no interesting output, e.g. `eslint: clean`). Not all of them have a golden test that actually exercises that path, so a regression in the empty-handling could go unnoticed.

### What to do
1. Find filters in [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters) that set `on_empty` but have no `[[tests.<name>]]` with an empty/clean `input`.
2. Add a golden test for each:
```toml
[[tests.<name>]]
name = "clean run collapses to message"
input = ""
expected = "<the on_empty text>"
```

### Verify
```bash
cargo test builtin_golden_tests_pass
```

### Acceptance criteria
- [ ] Every `on_empty`-bearing filter has a golden test for the empty input
- [ ] `cargo test builtin_golden_tests_pass` passes

AnayGarodia:
Closes #210.

## What

Adds a "clean run collapses to message" golden test to the eight filters that declared an `on_empty` message but had no test exercising the empty path: jest, mvn-build, mypy, pip, ruff, terraform-plan, tsc, vitest.

## Why

Without a test for the empty case, a regression in the empty-handling could silently stop emitting the `on_empty` message. The other 33 `on_empty` filters already have this coverage.

## How

Each new case feeds empty input and expects the filter's `on_empty` string, matching the existing pattern (e.g. `eslint`'s "clean run collapses to message"). For these keep-by-default filters, blank input is the only thing that reaches the `on_empty` branch, so it is the right exercise.

Verified with the golden-test runner (`cargo test builtin_golden_tests_pass`, and the `objects_e2e` integration check) — confirmed the new cases actually run by temporarily breaking one and seeing it caught.

AI-assisted.

### DIFF
--- a/assets/filters/jest.toml
+++ b/assets/filters/jest.toml
@@ -63,3 +63,8 @@ FAIL src/config.test.ts
 Test Suites: 1 failed, 1 passed, 2 total
 Tests:       3 passed, 3 total
 """
+
+[[tests.jest]]
+name = "clean run collapses to message"
+input = ""
+expected = "jest: ok"
--- a/assets/filters/mvn-build.toml
+++ b/assets/filters/mvn-build.toml
@@ -42,3 +42,8 @@ input = """
 [INFO] Finished at: 2024-01-15T10:30:00Z
 """
 expected = "[INFO] BUILD SUCCESS\n[INFO] Total time: 4.123 s\n[INFO] Finished at: 2024-01-15T10:30:00Z"
+
+[[tests.mvn-build]]
+name = "clean run collapses to message"
+input = ""
+expected = "mvn: ok"
--- a/assets/filters/mypy.toml
+++ b/assets/filters/mypy.toml
@@ -26,3 +26,8 @@ Found 2 errors in 2 files (checked 9 source files)
 name = "keeps success line"
 input = "Success: no issues found in 5 source files"
 expected = "Success: no issues found in 5 source files"
+
+[[tests.mypy]]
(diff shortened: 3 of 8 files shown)


Koukyosyumei (approved):
Reviewed and verified locally:

- **Expected strings all match**: each new case's `expected` is exactly the filter's declared `on_empty` message (checked all 8: `jest: ok`, `mvn: ok`, `mypy: ok`, `pip: ok`, `ruff: clean`, `terraform plan: no changes detected`, `tsc: ok`, `vitest: ok`).
- **The "remaining eight" claim holds**: I swept every `assets/filters/*.toml` with an `on_empty` — the others already exercise the empty path (either `input = ""` or an "empty after stripping" case like gradle/mise/task/turbo). These 8 were the only gaps.
- **Tests pass**: applied the diff and ran `cargo test --lib builtin_golden_tests_pass` — green.

One optional thought, not blocking: for the strip-based filters (e.g. jest), a "strips to empty" input (only `PASS`/noise lines, like gradle's `empty after stripping` case) would exercise the strip patterns *and* the `on_empty` branch together; blank input covers the branch itself, which is what #210 asked for. Fine as is.

LGTM.

PR #313, opened 2026-07-11
feat(filters): add an output filter for deno

### LINKED ISSUE #205: Add an output filter for `deno`
## Add an output filter for `deno`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time. This is a self-contained good-first-issue: **add a filter for `deno`**. No changes to core logic are needed.

### Why
`deno test`/`lint`/`check` print per-item progress; the signal is failures, diagnostics, and the test/lint summary.

### What to keep / strip
- **Keep (the signal):** failing test/lint names, error diagnostics, and the `N passed; M failed` summary.
- **Strip (the noise):** `ok`/passing lines, download notices, and blank lines.

### Steps
1. Create `assets/filters/deno.toml`.
2. Match the command with a `match_command` regex, e.g. `match_command = "^deno\\b"`. Make it match typical invocations but not unrelated commands.
4. Add **2–3 inline golden tests** (`[[tests.deno]]`): a normal run, an error/failure run, and a clean/empty run if relevant.

### Golden-test shape
```toml
[filters.deno]
description = "…"
match_command = "^deno\\b"
strip_ansi = true

[[tests.deno]]
input = """…raw output…"""
```

### How to verify
```bash
```

### Reference
The upstream [`rtk`](https://github.com/rtk-ai/rtk) project supports the Deno toolchain — see its `src/filters/` and `src/cmds/` for inspiration on what matters in the output. h5i's filters are derived from rtk (see `assets/filters/NOTICE`).

### Acceptance criteria
- [ ] `assets/filters/deno.toml` added with a clear `description`
- [ ] `match_command` matches typical `deno` invocations and not unrelated ones

AnayGarodia:
Closes #205.

## What

Adds `assets/filters/deno.toml` to compact `deno` output.

## Why

`deno test` / `lint` / `check` print per-item progress and download notices. The signal is failures, diagnostics, and the summary.

## How

Keep-by-default (strip only known noise) so an error, a diagnostic, or the summary is never dropped or turned into a false success. Strips passing (`... ok`) lines, `Download` / `Check` notices, `running N tests from ...`, and blank lines; keeps everything else. `on_empty = "deno: ok"` covers a fully-clean run.

Four golden tests from real `deno` output (captured with deno 2.9.2):
- a failing test run — keeps the `FAILED` line, the `ERRORS`/`FAILURES` sections, the assertion diff, the stack, and the `N passed | M failed` summary;
- an all-pass run — collapses to just the summary line;
- a `deno lint` run — keeps the rule diagnostics, location, hint, and docs link;
- a clean run — collapses to the `on_empty` message.

Reduction scales with suite size: passing-test spam is the bulk of a large run, so a big suite collapses to its failures plus the summary. `cargo test builtin_golden_tests_pass` and `cargo test --test filter_quality` pass.

AI-assisted; the golden inputs are real captured deno output, not synthetic (paths normalized).

### DIFF
--- a/assets/filters/deno.toml
+++ b/assets/filters/deno.toml
@@ -1,0 +1,107 @@
+[filters.deno]
+description = "Compact deno output — keep failures, diagnostics, and the summary; drop passing lines, download notices, and blank lines"
+match_command = "^deno\\b"
+strip_ansi = true
+# Keep-by-default (strip only known noise) so an unexpected error, a diagnostic,
+# or the summary is never dropped or turned into a false success.
+strip_lines_matching = [
+  "^\\s*$",
+  "^Download ",
+  "^Check ",
+  "\\.\\.\\. ok",
+  "^running \\d+ tests? from",
+]
+on_empty = "deno: ok"
+
+[[tests.deno]]
+name = "keeps failures, diagnostics, and the summary; drops passing/download noise"
+input = """
+Download https://jsr.io/@std/assert/1.0.19/assert.ts
+running 3 tests from ./math_test.ts
+adds numbers ... ok (523µs)
+multiplies numbers ... ok (73µs)
+this one fails ... FAILED (18ms)
+
+ ERRORS
+
+this one fails => ./math_test.ts:11:6
+error: AssertionError: Values are not equal.
+
+
+    [Diff] Actual / Expected
+
+
+-   2
++   3
+
+  throw new AssertionError(message);
+        ^
+    at assertEquals (https://jsr.io/@std/assert/1.0.19/equals.ts:67:9)
+    at file:///project/math_test.ts:12:3
+
+ FAILURES
+
+this one fails => ./math_test.ts:11:6
+
+FAILED | 2 passed | 1 failed (21ms)
+
+error: Test failed
+"""
+expected = """
+this one fails ... FAILED (18ms)
+ ERRORS
+this one fails => ./math_test.ts:11:6
+error: AssertionError: Values are not equal.
+    [Diff] Actual / Expected
+-   2
++   3
+  throw new AssertionError(message);
+        ^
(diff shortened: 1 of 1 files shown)


Koukyosyumei (approved):
Reviewed the diff and verified locally: applied the file on top of main and ran both `cargo test -p h5i-core builtin_golden_tests_pass` and `cargo test -p h5i-core --test filter_quality` — all pass, so the filter is picked up by the rust-embed auto-discovery with no registration needed.

**What it does:** adds a keep-by-default `deno` output filter (`assets/filters/deno.toml`) stripping passing-test lines, `Download`/`Check` notices, `running N tests from`, and blanks; four golden tests from real deno 2.9.2 output.

**Correctness / conventions:**
- `match_command = "^deno\b"` doesn't collide with any existing filter pattern (checked all 80 `match_command` regexes).
- Keep-by-default is the right call for a test runner — an unrecognized error line can never be turned into a false success, and the failing-run golden test proves the FAILED line, ERRORS/FAILURES sections, diff, stack, and summary all survive.
- The all-pass and `on_empty` collapses match the vitest/jest precedent.

**Non-blocking nits:**
- `"\.\.\. ok"` is an unanchored substring, so a user `console.log` like `warming cache... ok` inside a test would also be stripped. Anchoring on the duration suffix (`"\.\.\. ok \("`) would tighten it, since deno always prints one. Same class of tradeoff as `^Check ` — fine either way, keep-by-default limits the blast radius to single lines.
- No `max_lines` cap (vitest uses 80). For keep-by-default that's arguably intentional (never truncate a long failure), so just noting it.

LGTM. ✅

PR #314, opened 2026-07-11
refactor(workspace)

Koukyosyumei:

PR #315, opened 2026-07-11
feat(filters): add PHPUnit output filter

### LINKED ISSUE #204: Add an output filter for `phpunit`
## Add an output filter for `phpunit`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time. This is a self-contained good-first-issue: **add a filter for `phpunit`**. No changes to core logic are needed.

### Why
PHPUnit prints a progress line of dots/F/E characters; the signal is the failures block and the `Tests: N, Assertions: M, Failures: K` summary.

### What to keep / strip
- **Keep (the signal):** the `There was/were N failure(s)` block, failing test names, and the final summary.
- **Strip (the noise):** the per-test progress characters and blank lines.

### Steps
1. Create `assets/filters/phpunit.toml`.
2. Match the command with a `match_command` regex, e.g. `match_command = "\\bphpunit\\b"`. Make it match typical invocations but not unrelated commands.
4. Add **2–3 inline golden tests** (`[[tests.phpunit]]`): a normal run, an error/failure run, and a clean/empty run if relevant.

### Golden-test shape
```toml
[filters.phpunit]
description = "…"
match_command = "\\bphpunit\\b"
strip_ansi = true

[[tests.phpunit]]
input = """…raw output…"""
```

### How to verify
```bash
```

### Reference
The upstream [`rtk`](https://github.com/rtk-ai/rtk) project supports PHP testing — see its `src/filters/` and `src/cmds/` for inspiration on what matters in the output. h5i's filters are derived from rtk (see `assets/filters/NOTICE`).

### Acceptance criteria
- [ ] `assets/filters/phpunit.toml` added with a clear `description`
- [ ] `match_command` matches typical `phpunit` invocations and not unrelated ones

cnYui:
## Summary

- add a declarative PHPUnit output filter that removes blank lines and counted progress rows
- preserve headers, failure/error diagnostics, assertion diffs, stack traces, timing, and final summaries
- route direct PHPUnit commands plus common `php -d/-c/-n ... phpunit` invocations without claiming ordinary PHP scripts
- include three inline golden cases and explicit positive/negative command-routing coverage

Closes #204

## Why

PHPUnit's per-test progress rows consume context without helping an agent diagnose a run. This filter is intentionally keep-by-default: it removes only known progress/blank-line noise so multiple failure blocks and their final summary cannot be truncated.

Empty output is reported as `phpunit: no output` rather than a false success.

## Testing

Run on WSL/Linux with `H5I_SKIP_WEB_BUILD=1`, `--no-default-features`, and an isolated Cargo target directory:

- `cargo test builtin_golden_tests_pass` - passed
- `cargo test known_commands_route_to_expected_rules` - passed
- `cargo test phpunit_command_routing_stays_narrow` - passed
- `cargo test --test filter_quality` - 13/13 passed
- `cargo clippy --all-targets -- -D warnings` - passed
- `cargo build --all-targets` - passed
- `git diff --check` - passed

A broader no-default-features run passed the unit, CLI integration, and noun-workflow suites; seven existing env integration cases fail when the repository is executed from WSL's `/mnt/d` mount because their process-tier worktree paths cannot be statted.

The failure fixture is adapted from rtk's PHPUnit filter. Apache-2.0 attribution is covered by `assets/filters/NOTICE`.


Koukyosyumei (approved):
Reviewed and approving.

**What it does:** adds a declarative PHPUnit output filter (`assets/filters/phpunit.toml`) that strips only known noise (blank lines + counted progress rows like `..F... 110 / 110 (100%)`) while keeping headers, failure/error diagnostics, assertion diffs, stack traces, timing, and summaries — plus positive/negative command-routing tests in `filter_rules.rs`.

**Review notes:**
- Keep-by-default design is right for this engine: only two `strip_lines_matching` patterns, so multi-failure blocks and summaries can't be truncated. `on_empty = "phpunit: no output"` avoids a false success, consistent with the other rules.
- `match_command` is appropriately narrow: `phpunit`/`phpunit.phar` directly, or `php` with `-n`/`-d val`/`-c val` options followed by a path ending in `phpunit(.phar)` — and the negative tests (`php script.php phpunit`, `php -r`, `phpunit-wrapper`) pin that down. One known miss: the attached-value form `php -dmemory_limit=-1 vendor/bin/phpunit` won't route (the regex requires a space after `-d`/`-c`). That fails open (output just passes through unfiltered), so fine to leave or tighten later.
- Schema, golden-test format, and rtk attribution (`assets/filters/NOTICE`) all match the existing filter set; the TOML self-registers via the directory embed and its golden cases run under `builtin_golden_tests_pass`.

**Conflict resolution:** the branch was based just before the workspace refactor (#314) moved `src/filter_rules.rs` → `crates/h5i-core/src/filter_rules.rs`, which is all GitHub was flagging. I merged `main` into the branch (git's rename detection carried the test hunks to the new path; the TOML merged cleanly) and pushed the merge commit.

**Verified on the merged head:** `cargo test -p h5i-core --lib filter_rules` — 9/9 passed (golden tests, routing, name-sort/ambiguity guard) — and `cargo clippy -p h5i-core --lib --tests -- -D warnings` is clean.

PR #327, opened 2026-07-12
feat(msg): add JSON inbox and history output

### LINKED ISSUE #324: Add `--json` to `h5i msg inbox` and `h5i msg history`
## Add `--json` to `h5i msg inbox` / `h5i msg history`

The i5h message channel has **no machine-readable output at all**: every `h5i msg` view is human-formatted text. An agent or script that wants to consume its inbox programmatically (e.g. route ASK vs REVIEW messages, or feed a dashboard) has to scrape the pretty view — which is also `sanitize_display`-escaped for terminals, so scraping doesn't even recover the original fields. The underlying `Message` type (`crates/h5i-core/src/msg.rs`) is already a serde struct persisted as `messages.jsonl`, so this is exposure, not new plumbing.

### Where
- CLI: `src/cli/msg.rs` — `MsgCommands::Inbox` (around line 168) and `MsgCommands::History` (around line 178), plus their match arms (around lines 522 and 555).
- Core: `inbox(...)` / the history read path in `crates/h5i-core/src/msg.rs` already return the structured `Message` records the text view renders.

### What to do
1. Add `#[arg(long)] json: bool` to both variants.
2. When set, serialize the same `Vec<Message>` the text view renders with `serde_json::to_string_pretty` (match the `env … --json` handlers).
3. Keep the read-state semantics identical: `inbox --json` still advances the cursor exactly like `inbox` (and `--peek --json` still doesn't). JSON changes the rendering, never the side effects.
4. **Do not sanitize the JSON fields** — emit the raw stored values, and note in the flag's help text that pulled message fields are untrusted collaborator input (`sanitize_display` is a terminal-injection defense for the *text* view; JSON consumers get the raw record and must treat it as data).

### How to verify
```bash
h5i msg send claude "hello" --from codex
h5i msg inbox --peek --json | jq .      # valid JSON, cursor not advanced
h5i msg inbox --json | jq '.[].body'    # valid JSON, cursor advanced
h5i msg history --json | jq length
h5i msg inbox                           # text view unchanged
```

### Acceptance criteria
- [ ] `h5i msg inbox --json` and `h5i msg history --json` emit valid JSON of the message records
- [ ] Cursor/read-state behavior is identical to the text views (including `--peek`)
- [ ] Text output unchanged when the flag is absent
- [ ] A test covers the JSON shape (see `crates/h5i-core/src/msg.rs` tests for fixtures, or a CLI test in `tests/`)

Bortlesboat:
## Summary
- add raw structured JSON output to msg inbox and history
- preserve inbox cursor and peek behavior in JSON mode
- document both flags and cover raw fields, empty output, and cursor semantics

Closes #324

## Verification
- H5I_SKIP_WEB_BUILD=1 cargo test --test msg_integration --no-run
- H5I_SKIP_WEB_BUILD=1 cargo clippy --bin h5i -- -D warnings
- git diff --check

The JSON integration test compiles on Windows. Running the harness locally is blocked by an existing Windows stack overflow during its h5i init fixture; the repository CI can execute it on its supported runners.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -2802,7 +2802,7 @@ h5i msg setup [<name>] [--scope project|user] [--no-block]
 h5i msg send <agent> <text>               # `all` = broadcast
 h5i msg ask|review|risk|handoff <agent> <text> [flags]
 h5i msg reply|ack|done|decline <n> [text]
-h5i msg inbox [--peek] | history [--with <agent>] | replay [--with <agent>] [--interval S] | team
+h5i msg inbox [--peek] [--json] | history [--with <agent>] [--json] | replay [--with <agent>] [--interval S] | team
 h5i msg wait [--all] [--timeout N] | watch [--all] | hook [--block] | as <name> | whoami
 ```
 
@@ -2852,10 +2852,12 @@ structured fields. **Options must precede the recipient** (the body is variadic)
 ```
 h5i msg                          # dashboard: header · inbox · GIT PROOF band (a glance; does not consume)
 h5i msg inbox                    # show unread, mark read, number them
+h5i msg inbox --peek --json      # raw unread message records; do not mark read
 h5i msg reply 1 on it            # threaded reply to message #1 of your last view
 h5i msg ack 1                    # ACK / DONE / DECLINE are typed threaded replies
 h5i msg done 1 fixed in 1a2b3c4
 h5i msg history --with codex     # full conversation log
+h5i msg history --json           # machine-readable message history
 h5i msg replay --with codex      # replay the log as a live feed (1s between messages)
 h5i msg team                     # known agents
 ```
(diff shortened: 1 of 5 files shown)


Koukyosyumei (approved):
Reviewed the diff and verified locally on Linux (since your Windows run was blocked): applied the patch on main, ran the **full** `msg_integration` suite (40/40 pass, including the new `inbox_and_history_emit_raw_json_without_changing_cursor_semantics`), `cargo clippy --all-targets --all-features -- -D warnings` clean, plus a live CLI smoke test (`inbox --peek --json` emits clean raw records).

**Matches every acceptance criterion of #324:**
- Cursor semantics are untouched: `--json` only changes rendering — `inbox --json` still advances the cursor, `--peek --json` doesn't, and `write_last_view` still runs so `reply <n>` works after a JSON read. The test proves the drain (second `inbox --json` returns `[]`).
- Raw fields, no `sanitize_display` — exactly right for machine consumers, and the test's `<message>` body pins that down. The "untrusted collaborator input" note in the flag help is the right place for the caveat.
- Empty result is `[]`, not a prose message — correct for JSON mode.
- `serde_json::to_string_pretty` matches the `env … --json` convention; docs updated in all three places (MANUAL.md, HTML, man page).

LGTM. ✅

PR #328, opened 2026-07-12
feat(msg): warn about unknown recipients

### LINKED ISSUE #325: `h5i msg send`: warn when the recipient has never been seen in the roster
## Warn when sending to an agent the roster has never seen

`h5i msg send codx "please review"` (note the typo) succeeds silently today: the message is appended, the roster even **adds** `codx` as a known agent, and the real `codex` never sees it. Nothing ever surfaces the mistake — the sender just waits forever for a reply. A one-line warning at send time ("never seen before, known agents are: …") catches the typo immediately while still allowing legitimate first-contact sends.

### Where
- Core: `crates/h5i-core/src/msg.rs` — `send_msg` (around line 291); the roster helpers `known_agents` / `Roster` (around line 199). Note the roster **auto-inserts the recipient** during append (around line 444), so the check must read the roster **before** the send is committed.
- CLI: `src/cli/msg.rs` — the `send`/`ask`/`review`/`risk`/`handoff` arms all funnel into the same core send; a single warning in the shared path covers them all.

### What to do
1. Before appending, check whether `to` is already in the roster (skip the check for the `all` broadcast and for self-sends).
2. If unknown, print a warning to **stderr** (stdout may be piped) and still send — first contact with a genuinely new agent is a supported flow, so this must not become an error:
   ```
   note: 'codx' has never been seen in this clone; known agents: claude, codex
   ```
3. When the roster is empty or only contains the sender (a fresh solo clone), stay quiet — warning on the very first message of a new clone is pure noise.

### How to verify
```bash
h5i msg send codex "hi" --from claude     # seeds the roster
h5i msg send codx "typo" --from claude    # warns on stderr, still sends
h5i msg send all "broadcast" --from claude  # no warning
```

### Acceptance criteria
- [ ] Sending to a name not in the roster prints a stderr warning listing known agents
- [ ] The message is still sent (warning, not error) and the exit code stays 0
- [ ] No warning for `all`, for known agents, or in a fresh clone with no roster yet
- [ ] Unit test alongside the existing `msg.rs` tests (the fixtures there make roster setup easy)

Bortlesboat:
## Summary
- inspect the persisted roster before appending a message
- warn on stderr for an unknown recipient while still sending successfully
- stay quiet for broadcasts, self-sends, known recipients, and fresh solo clones
- cover the roster and warning behavior with a focused unit test

Closes #325

## Verification
- H5I_SKIP_WEB_BUILD=1 cargo clippy -p h5i-core --lib --all-features -- -D warnings
- H5I_SKIP_WEB_BUILD=1 cargo build -p h5i-core
- git diff --check

The focused lib-test target is currently blocked on Windows by an existing unrelated env.rs test compile error: RunLock is Unix-gated but referenced by an ungated test.

### DIFF
--- a/crates/h5i-core/src/msg.rs
+++ b/crates/h5i-core/src/msg.rs
@@ -299,6 +299,10 @@ pub fn send_msg(
     validate_name(from)?;
     validate_name(to)?;
 
+    if let Some(warning) = unknown_recipient_warning(repo, from, to) {
+        eprintln!("{warning}");
+    }
+
     let ts = now_ts();
     let id = gen_id(&ts, from, to, body);
     let tag = opts.tag.map(|t| t.trim().to_string()).filter(|t| !t.is_empty());
@@ -353,6 +357,24 @@ pub fn send_msg(
         write_identity(h5i_root, from)?;
     }
     Ok(msg)
+}
+
+fn unknown_recipient_warning(repo: &Repository, from: &str, to: &str) -> Option<String> {
+    if to == BROADCAST || to == from {
+        return None;
+    }
+
+    let roster = read_roster(repo);
+    let known: Vec<&str> = roster.agents.keys().map(String::as_str).collect();
+    let has_other_agent = roster.agents.keys().any(|name| name != from);
+    if !has_other_agent || roster.agents.contains_key(to) {
+        return None;
+    }
+
+    Some(format!(
+        "note: '{to}' has never been seen in this clone; known agents: {}",
+        known.join(", ")
+    ))
 }
 
 /// Send a threaded reply to `original`, as `me`.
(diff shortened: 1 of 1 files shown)


Koukyosyumei (approved):
Reviewed the diff and verified locally on Linux (since your Windows lib-test run was blocked): the new unit test passes, the **full** `msg_integration` suite still passes (40/40 — confirms the new stderr output breaks no existing CLI expectations), clippy `-D warnings` clean, and a live CLI smoke test behaves exactly as specced: `msg send codx "typo"` prints `note: 'codx' has never been seen in this clone; known agents: claude, codex` on stderr and still exits 0.

**Matches every acceptance criterion of #325**, with the details handled correctly:
- The roster is read **before** `append_message_cas`, so the check isn't defeated by the roster's own auto-insert of the recipient.
- `Roster.agents` is a `BTreeMap`, so the exact-string assertion in the test is deterministic — I checked this specifically since a `HashMap` would have made it flaky.
- Quiet cases all covered: broadcast, self-send, known recipient, and fresh solo clone (`has_other_agent` handles both the empty roster and the sender-only roster in one predicate).
- Placing the check in `send_msg` covers `send`/`ask`/`review`/`risk`/`handoff` in one spot, and the reply path can't spuriously warn since a reply's recipient is by construction already in the roster.

**Non-blocking nit:** `eprintln!` from `h5i-core` is a small layering smell (a library writing to stderr directly) — a future refactor could return the warning for the CLI to print. Not worth blocking on: stderr is safe in every current consumer, including MCP stdio mode where stdout is the protocol channel.

LGTM. ✅

PR #329, opened 2026-07-12
feat: h5i env rm accepts multiple names

### LINKED ISSUE #326: `h5i env rm`: accept multiple environment names
## Let `h5i env rm` remove several environments in one call

The arena workflow creates environments in batches — `h5i env create` N variants, `h5i env compare` them, apply the winner. Cleaning up the losers is then one `h5i env rm` **per env**, each repeating the same lookup and confirmation ceremony. `h5i env rm exp1 exp2 exp3` should just work, like `git branch -d a b c`.

### Where
- CLI: `src/cli/env.rs` — the `EnvCommands::Rm` variant (around line 238, currently `Rm { name: String, force: bool }`) and its match arm (around line 929). The removal logic itself lives in `h5i_core::env` and needs no changes — this is a CLI-surface change.

### What to do
1. Change `name: String` to `names: Vec<String>` with `#[arg(required = true, num_args = 1..)]`.
2. Loop the existing single-env removal over the names. **Keep going on error** (report the failure, continue with the rest — one missing env must not strand the others un-removed), and exit non-zero at the end if any removal failed.
3. `--force` applies to every listed env, matching flag semantics elsewhere.
4. Update the `env rm` row in `MANUAL.md` and regenerate the man page/HTML manual (`scripts/gen_man.sh`, `python3 scripts/gen_manual.py`) — see #306 for what a docs-regeneration diff looks like.

### How to verify
```bash
h5i env create a && h5i env create b
h5i env rm a b --force            # both gone in one call
h5i env rm nope also-nope; echo $?  # reports each failure, exits non-zero
h5i env list                      # neither a nor b remains
```

### Acceptance criteria
- [ ] `h5i env rm <n1> <n2> …` removes every named env; single-name behavior unchanged
- [ ] A failing name doesn't abort the rest; the exit code is non-zero if any failed
- [ ] Docs regenerated (MANUAL.md, man page, HTML manual)
- [ ] Integration test in `tests/env_integration.rs` covering the multi-name and partial-failure paths

LevelVoid:
Closes #326

Acceptance criteria

- [x] h5i env rm <n1> <n2> … removes every named env; single-name behavior unchanged

- [x] A failing name doesn't abort the rest; the exit code is non-zero if any failed

- [x] Docs regenerated (MANUAL.md, man page, HTML manual)

- [x] Integration test in tests/env_integration.rs covering the multi-name and partial-failure paths

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -2932,6 +2932,7 @@ another reviews and applies). See `docs/environments-design.md` and the live
 | `h5i env propose <name>` | Mediated commit (path-allowlist enforced: rejects nested `.git`, symlink escapes, `..`) + review brief. Never writes the parent. |
 | `h5i env apply <name> [--patch]` | Apply a proposed env onto its parent (reviewer-selected). Default merges; `--patch` squashes into one commit. The applied commit is **stamped with env provenance** — a note linking it back to the env and summarizing the evidence by trust lane (`host-env-run` vs box-claimed `inbox-capture`), visible in `h5i recall log`. |
 | `h5i env abort <name>` | Discard the env; manifest + workspace retained for forensics. |
+| `h5i env rm <name>… [--force]` | **Permanently** remove one or more environments: prune their worktrees, delete their code + reasoning branches, and erase their manifests. Multiple names are processed in order; a failure for one does not abort the rest, but the exit code is non-zero if any removal failed. A still-live env (created/running/proposed) needs `--force`. Only the `removed` event in `refs/h5i/env` survives. |
 | `h5i env gc` | Reclaim worktrees of applied/aborted envs. Manifests, branches, and captures are retained. |
 
 `<name>` accepts a bare slug, `agent/slug`, or the full `env/agent/slug`.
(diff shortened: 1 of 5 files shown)


Koukyosyumei (approved):
Reviewed the diff and verified locally: applied the patch on main, ran both new integration tests (`env_rm_removes_multiple_envs`, `env_rm_partial_failure_continues_and_exits_nonzero` — pass), and `cargo clippy --all-targets --all-features -- -D warnings` is clean.

**Matches every acceptance criterion of #326:**
- `#[arg(required = true, num_args = 1..)]` keeps the single-name invocation byte-identical while accepting N names.
- Keep-going-on-error is implemented at both failure points (`find` and `rm`), each reported to stderr with the offending name, and the non-zero exit only fires at the end — the partial-failure test proves the surviving env is still removed.
- `std::process::exit(1)` matches the existing CLI convention (`cli/commit.rs`, `cli/resolve.rs`, `cli/objects.rs` do the same), and there's no cleanup after the match arm for it to skip.
- Docs regenerated in all three places, and the man-page `NAMES` entry even gained a description the old `NAME` lacked.
- The tests use the existing `Repo` harness helpers (`env_dir`/`work`/`h5i` vs `h5i_ok`) idiomatically, asserting on manifests, worktrees, `env list`, exit code, and both output streams.

**Trivia, not blocking:** a duplicated name (`h5i env rm a a --force`) removes on the first pass, fails "not found" on the second, and exits non-zero — same behavior as `git branch -d a a`, so arguably correct as-is.

LGTM. ✅

PR #331, opened 2026-07-12
feat(env): add limit to event log

### LINKED ISSUE #323: Add `--limit` to `h5i env log` to tail long event logs
## Add `--limit N` to `h5i env log`

`h5i env log <name>` prints the env's **entire** event log. A long-lived env accumulates an `exec` event for every `env run`/`env shell` command, so after a real working session the interesting part (the most recent lifecycle events) scrolls far off screen. `h5i recall log` already has `--limit` (default 10, `0` = all) — `env log` should offer the same tail.

### Where
- CLI: `src/cli/env.rs` — the `EnvCommands::Log` variant (around line 156) and its match arm in `run()` (around line 812). Events come from `h5i_core::env::read_events(git, Some(&m.id))`, which returns a `Vec<EnvEvent>` ordered oldest-first.

### What to do
1. Add `#[arg(long, default_value_t = 0)] limit: usize` to the `Log` variant — `0` = all, matching the `h5i recall log` convention. Keeping the default at "all" preserves today's behavior.
2. In the handler, when `limit > 0`, keep only the **last** N events (it's a tail, not a head — the newest events are the signal).
3. #306 adds `--json` to the same subcommand; `--limit` must compose with it (apply the tail before serializing).

### How to verify
```bash
h5i env create demo
h5i env run demo -- true && h5i env run demo -- true && h5i env run demo -- true
h5i env log demo --limit 2          # only the 2 most recent events
h5i env log demo                    # unchanged: full log
h5i env log demo --limit 2 --json   # JSON view respects the limit (after #306 lands)
```

### Acceptance criteria
- [ ] `h5i env log <name> --limit N` shows only the N most recent events
- [ ] `--limit 0` (the default) shows everything — current behavior unchanged
- [ ] Composes with `--json` (#306)
- [ ] Covered by an integration test in `tests/env_integration.rs`

Bortlesboat:
## Summary

- add `--limit N` to `h5i env log`, preserving the current unlimited default
- retain the newest N events rather than the oldest N
- apply the tail before both text rendering and JSON serialization

Closes #323.

## Validation

- `H5I_SKIP_WEB_BUILD=1 cargo check --bin h5i`
- `H5I_SKIP_WEB_BUILD=1 cargo clippy --bin h5i -- -D warnings`
- `git diff --check`
- added a Linux integration regression covering newest-event ordering, JSON composition, and `--limit 0`

The full integration target cannot compile on native Windows because existing tests import `std::os::unix::fs::PermissionsExt` unconditionally. A native CLI smoke test is also blocked by a pre-existing stack overflow in the freshly built Windows binary. Upstream Linux CI will run the added end-to-end test.

### DIFF
--- a/man/man1/h5i.1
+++ b/man/man1/h5i.1
@@ -518,10 +518,13 @@ Show the event log (refs/h5i/env) for one environment
 .ie \n(.g .ds Aq \(aq
 .el .ds Aq '
 .SS SYNOPSIS
-\fBlog\fR [\fB\-\-json\fR] [\fB\-h\fR|\fB\-\-help\fR] <\fINAME\fR> 
+\fBlog\fR [\fB\-\-limit\fR] [\fB\-\-json\fR] [\fB\-h\fR|\fB\-\-help\fR] <\fINAME\fR> 
 .ie \n(.g .ds Aq \(aq
 .el .ds Aq '
 .SS OPTIONS
+.TP
+\fB\-\-limit\fR \fI<LIMIT>\fR [default: 0]
+Show only the newest N events (0 shows the full log)
 .TP
 \fB\-\-json\fR
 Emit the event records as JSON instead of the human view
--- a/src/cli/env.rs
+++ b/src/cli/env.rs
@@ -164,6 +164,9 @@ pub enum EnvCommands {
     /// Show the event log (refs/h5i/env) for one environment
     Log {
         name: String,
+        /// Show only the newest N events (0 shows the full log)
+        #[arg(long, default_value_t = 0)]
+        limit: usize,
         /// Emit the event records as JSON instead of the human view
         #[arg(long)]
         json: bool,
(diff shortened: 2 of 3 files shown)


Koukyosyumei (approved):
## Review

### Overview
Adds `--limit N` to `h5i env log`, tailing to the newest N events (default `0` = full log, preserving current behavior), applied before both the human view and `--json` serialization. Closes #323.

### Correctness — verified
- `env::read_events` reads the append-only `events.jsonl` in file order (oldest-first), so `events.drain(..events.len() - limit)` keeps exactly the **newest** N — a tail, as #323 asks. The `events.len() > limit` guard rules out underflow.
- The tail is applied once, before the `json`/text branch, so the two views can't drift.
- `0` = unlimited matches the existing `h5i recall log` sentinel convention.
- Man page: the diff matches `gen_man.sh` output exactly (synopsis + OPTIONS style), and the "manuals fresh" CI check confirms it's regenerated, not hand-edited.

### Test coverage
`env_log_limit_returns_the_newest_events` is well-constructed: it asserts the limited view equals the **tail** of the full view (catches an accidental head-N regression), covers `--limit 0` ≡ unlimited, and runs through the `Repo` helper's `H5I_DEFAULT_ISOLATION=workspace` pin, so it's portable in CI with no sandbox capability requirements.

### Nits (non-blocking)
- Consider a `-l` short flag to mirror `h5i recall log -l/--limit` and `h5i msg history -l`. Fine as a follow-up (would need a man regen).

All acceptance criteria of #323 are met (newest-N, default unchanged, composes with `--json`, integration-tested). CI is green across build/test/clippy/fmt, cross-checks, and manuals. LGTM.

PR #332, opened 2026-07-12
feat(log): add JSON output

### LINKED ISSUE #322: Add `--json` to `h5i recall log`
## Add `--json` to `h5i recall log` for machine-readable provenance history

`h5i recall log` is the main window into the enriched 5D commit history (message, AI model/agent, prompt, token counts, test metrics), but it only has a human-formatted view. The `env` family already exposes `--json` everywhere (`env list/status/diff/inspect/compare/capabilities/ports`, and `env log` in #306), so tooling that wants the provenance feed has to scrape text. The records are already structured (`H5iCommitRecord` in `crates/h5i-core/src/metadata.rs` is a serde type), so this is exposure, not new plumbing.

### Where
- CLI: `src/cli/log.rs` — `run(limit, ancestry)` is the handler for `h5i recall log` (and the hidden legacy `h5i log`); the clap variant is `Commands::Log` in `src/main.rs` (around line 843).
- Core: `crates/h5i-core/src/repository.rs` — `print_log(limit)` (around line 585) walks HEAD and prints each commit's record inline.

### What to do
1. Add a `#[arg(long)] json: bool` to the `Log` variant and thread it through `cli::log::run`.
2. `print_log` formats while it walks, so extract the record-collection step: a method like `log_records(limit) -> Vec<…>` that both the pretty view and the JSON path share (each entry: commit SHA + the attached `H5iCommitRecord`, if any). Keep `print_log`'s output byte-identical.
3. Serialize with `serde_json::to_string_pretty`, matching the `env … --json` handlers.

### How to verify
```bash
h5i recall log --limit 5 --json | jq .   # valid JSON array of commit records
h5i recall log --limit 5                 # human view unchanged
```

### Acceptance criteria
- [ ] `h5i recall log --json` emits valid JSON (SHA + provenance record per commit)
- [ ] `--limit` applies to the JSON view too
- [ ] The default text output is unchanged
- [ ] An integration test in `tests/` parses the output with `serde_json` (see `log_json_emits_env_event_records` in `tests/env_integration.rs` for the pattern)

Bortlesboat:
## Summary

- add `--json` to `h5i recall log` and the legacy `h5i log` alias
- serialize the existing enriched commit records, including `git_oid` and provenance metadata
- preserve the current human-readable output and apply `--limit` identically to both views
- reject the incompatible `--json --ancestry` combination at argument parsing

Closes #322.

## Validation

- Linux: `H5I_SKIP_WEB_BUILD=1 cargo test --test e2e_noun_workflow log_` — 3 passed
- Windows: `H5I_SKIP_WEB_BUILD=1 cargo clippy --bin h5i -- -D warnings`
- Windows: `H5I_SKIP_WEB_BUILD=1 cargo check --bin h5i`
- `git diff --check` on changed source/test files

The regression compiles on Windows, but execution there remains blocked by the repository's pre-existing startup stack overflow. The same end-to-end test passes under Linux, matching upstream CI.

### DIFF
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -1,7 +1,7 @@
 //! `h5i log` — CLI handler (migrated from main.rs).
 use crate::*;
 
-pub fn run(limit: usize, ancestry: Option<String>) -> anyhow::Result<()> {
+pub fn run(limit: usize, json: bool, ancestry: Option<String>) -> anyhow::Result<()> {
     {
             let repo = H5iRepository::open(".")?;
 
@@ -72,7 +72,11 @@ pub fn run(limit: usize, ancestry: Option<String>) -> anyhow::Result<()> {
                 }
             } else {
                 let log_limit = if limit == 0 { usize::MAX } else { limit };
-                repo.print_log(log_limit)?;
+                if json {
+                    println!("{}", serde_json::to_string_pretty(&repo.get_log(log_limit)?)?);
+                } else {
+                    repo.print_log(log_limit)?;
+                }
             }
         }
     Ok(())
--- a/src/main.rs
+++ b/src/main.rs
@@ -847,6 +847,10 @@ enum Commands {
         #[arg(short, long, default_value_t = 10)]
         limit: usize,
 
+        /// Emit commit provenance records as JSON
+        #[arg(long, conflicts_with = "ancestry")]
+        json: bool,
+
         /// Show the full prompt ancestry chain for a specific line.
         /// Format: <file>:<line>  e.g.  src/model.py:42
         /// Prints every commit that ever touched that line, annotated with the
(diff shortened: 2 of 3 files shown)


Koukyosyumei (approved):
## Review

### Overview
Adds `--json` to `h5i recall log` (and the legacy `h5i log` alias), serializing the enriched `H5iCommitRecord` provenance feed with `--limit` applied identically to both views. Closes #322.

### Correctness — verified
- Reusing the existing `H5iRepository::get_log(limit)` instead of adding a new collection method is the right call — it walks HEAD with the same revwalk as `print_log`, so ordering and `limit` semantics (incl. the `0 → usize::MAX` sentinel) match the human view exactly, and the text output stays byte-identical as #322 requires.
- `get_log`'s `minimal_from_git` fallback means plain (non-h5i) commits appear as records with `ai_metadata: null` rather than being silently dropped — good behavior for a machine feed.
- Because the `recall` noun is rewritten pre-clap onto `Commands::Log`, the single flag covers both `h5i recall log --json` and legacy `h5i log --json`.
- `conflicts_with = "ancestry"` rejects the meaningless combination at parse time instead of silently ignoring the flag — nice fail-fast touch.
- `serde_json::to_string_pretty` matches the established `env … --json` handlers.
- No man-page regen needed: the legacy `log` verb is hidden from the generated man page (confirmed by the green "manuals fresh" check).

### Test coverage
`recall_log_json_emits_limited_commit_records` parses the output with `serde_json` and asserts `--limit`, `git_oid`, and `ai_metadata` presence — matches the pattern #322 points at.

### Nits (non-blocking)
- The static `recall` noun-help example (`h5i recall log --limit 20` in `main.rs`) could mention `--json` for discoverability; fine as a follow-up.

All acceptance criteria of #322 are met. CI is green across build/test/clippy/fmt, cross-checks, and manuals. LGTM.

PR #334, opened 2026-07-13
fix(sandbox): mount private tmp after home state

### LINKED ISSUE #320: agent profile fails with EINVAL when repository is under /tmp
## Summary

An `agent-claude` environment created from a repository located under `/tmp` cannot run even `/bin/true`. The same commands succeed when the repository is located under `$HOME`, with the same h5i binary and resolved policy digest.

## Environment

- h5i: `0.2.8`
- OS: WSL2 Linux `6.6.87.2-microsoft-standard-WSL2`
- Architecture: `aarch64`
- Landlock ABI: 3
- User namespaces: available
- Seccomp user notification: reported available

`h5i env capabilities` reports `supervised` as satisfiable and runnable.

### Repository under `/tmp` (fails)

```bash
repo=$(mktemp -d /tmp/h5i-agent-claude.XXXXXX)
git -C "$repo" init -q
git -C "$repo" config user.name test
git -C "$repo" config user.email test@example.invalid
touch "$repo/README.md"
git -C "$repo" add README.md
git -C "$repo" commit -qm init
cd "$repo"

h5i env create claude-env --profile agent-claude
h5i env run claude-env -- /bin/true
```

Result:

```text
Created environment env/human/claude-env (isolation: supervised, profile: agent-claude)
Error: Metadata error: supervised spawn failed: Invalid argument (os error 22)
```

### Repository under `$HOME` (succeeds)

Repeat the same steps, changing only the first line:

```bash
repo=$(mktemp -d "$HOME/h5i-agent-claude.XXXXXX")
```

Result:

```text
true generic ok
evidence e3b0c44298fc1c14 (env env/human/claude-env, policy eb7d2b04fb47) · wall 44ms, cpu 0ms, rss 0MiB
```

The resolved policy digest is `eb7d2b04fb47` in both cases.

## Additional isolation

- Launching the failing `/tmp` command through a user `systemd-run` service produces the same error, so it is not dependent on the original caller or process ancestry.
- A `default`-profile environment can run `/bin/true`; the failure is associated with the agent-profile setup.
- The same location-dependent failure was seen after overriding `agent-claude` to `isolation = "process"`, `net.mode = "host"`, and `net.egress = []`, suggesting this is not specific to supervised networking.

## Suspected area

This may involve overlapping filesystem rules or bind mounts: the built-in agent profile grants `/tmp` writable, while `$WORK` is nested beneath `/tmp` in the failing case. This is only a hypothesis; the current error does not identify which `pre_exec` operation returned `EINVAL`.

It would also help if the spawn error included the failing setup stage or syscall so this class of failure is easier to diagnose.

Bortlesboat:
## Summary

- mount the private `/tmp` bind after HOME-state bind sources are established
- preserve the existing bind set while making mount ordering explicit
- add a regression for repositories under `/tmp`

Closes #320.

## Validation

- exact WSL process-tier reproduction from a repository under `/tmp`
- `h5i-sandbox`: 170 tests passed
- `h5i-core`: 846 passed, 3 ignored
- Windows Clippy for both libraries with `-D warnings`
- `git diff --check`

### DIFF
--- a/crates/h5i-sandbox/src/sandbox.rs
+++ b/crates/h5i-sandbox/src/sandbox.rs
@@ -106,6 +106,16 @@ fn config_lock_paths(work: &Path, home: Option<&Path>) -> Vec<PathBuf> {
     out
 }
 
+/// Apply the private `/tmp` bind last. Its backing and the agent HOME-state
+/// copies can live below a repository in `/tmp`; mounting it first hides those
+/// sources before their own binds run.
+#[cfg(target_os = "linux")]
+fn home_binds_in_mount_order(binds: &[HomeBind]) -> Vec<&HomeBind> {
+    let mut ordered: Vec<_> = binds.iter().collect();
+    ordered.sort_by_key(|bind| bind.target == Path::new("/tmp"));
+    ordered
+}
+
 // ── raw TOML schema (what users write; everything optional) ────────────────
 
 #[derive(Debug, Default, Deserialize)]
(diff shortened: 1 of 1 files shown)


Koukyosyumei:
Thank you so much for your contribution and help! We’ve decided to put `h5i env` on hold for the time being, so I’ll close this PR for now.

PR #343, opened 2026-07-14
feat(env): add JSON allowlist output

### LINKED ISSUE #339: Add `--json` to the `h5i env allow` list view
## Summary

`h5i env allow` with no rule lists the persistent user-level egress allowlist (the extra hosts merged into container-tier envs). The env family is otherwise fully scriptable (`env list --json`, `env status --json`, `env capabilities --json`, `env compare --json`), but the allowlist can only be read as styled text, so tooling that wants to audit "what egress has this user persistently granted" has to parse it.

## Proposed behavior

```bash
h5i env allow --json
```

emits a JSON document with the stored rules and the file they come from, e.g.:

```json
{
  "path": "/home/user/.config/h5i/allow.toml",
  "rules": [".crates.io", "api.github.com:443"]
}
```

`--json` only applies to the list form; combining it with a rule argument or `--remove` should be a clear error (or simply be rejected by clap with `conflicts_with`).

## Where to look

- Handler: `src/cli/env.rs`, the `EnvCommands::Allow` arm around line 525. The list branch already calls `h5i_core::env::user_allow_list()` and `h5i_core::env::user_allow_path()`, which is exactly the data the JSON needs.
- Clap: the `Allow` variant at `src/cli/env.rs:89`; use `conflicts_with = "remove"` and make the flag require the no-rule form.
- Precedent: any of the existing env `--json` arms in the same file for output conventions (print with `serde_json::to_string_pretty`, no styling).

Note that `env allow` refuses to run inside a box (host-side only); the `--json` path should keep that behavior.

## Acceptance criteria

- `h5i env allow --json` prints valid JSON only; an empty allowlist yields `"rules": []`.
- `h5i env allow <rule> --json` and `--remove --json` are rejected with a helpful message.
- Default styled listing and the add/remove flows are unchanged.
- A test asserting the JSON shape after adding one rule (an integration test can point `$XDG_CONFIG_HOME`/`$HOME` at a temp dir the way existing env tests do).

ychampion:
## Summary

- add a list-only `h5i env allow --json` view with the allowlist path and normalized rules
- reject JSON with add/remove forms and preserve host-only policy boundaries
- cover stored, empty, conflicting, and in-box cases

## Why

The env command family is otherwise scriptable, but auditing persistent egress grants requires parsing styled text. This adds stable JSON without changing add/remove or default listing behavior.

Closes #339.

## Validation

- `cargo test --test env_integration env_allow_add_list_remove_and_in_box_refusal -- --nocapture` — passed
- `cargo build --all-targets` — passed
- `./scripts/gen_man.sh` — passed
- `python3 scripts/gen_manual.py` — passed


Koukyosyumei:

PR #355, opened 2026-07-19
docs: benchmark env isolation overhead

### LINKED ISSUE #353: Benchmark: `h5i env run` sandbox overhead per isolation tier
## Summary

`h5i env run` executes a command inside a sandboxed worktree with policy enforcement (Landlock + seccomp on the kernel tiers, rootless Podman on the container tier). Users deciding whether to run their build/test loop inside an env keep asking the same question: **how much slower is it than running bare?** We have no published numbers.

This task is to benchmark the sandbox overhead per isolation tier on your machine and check the methodology + results into the docs. It needs **no Rust knowledge**, just actually using `h5i env`.

## What to do

1. Install h5i and run `h5i env probe` and `h5i env capabilities --json` to see which tiers your host supports (`workspace`, `process`, `supervised`, `container`). Benchmark only the tiers that probe green; that partial coverage is fine and expected.
2. Pick 2 fixed workloads, one short and one long, for example:
   - short: `git status` or `python3 -c "print(1)"` (measures pure per-run setup cost),
   - long: `cargo build` on a small crate, or a real test suite (measures steady-state overhead).
3. For each supported tier:
   ```bash
   h5i env create bench-<tier> --isolation <tier>
   h5i env run bench-<tier> -- <workload>     # repeat 5x
   ```
   and compare against 5 bare runs of the same workload in a plain worktree.
4. Collect timings. `h5i env log <name>` already records wall/cpu/peak-RSS accounting per exec event, so you can pull the sandboxed numbers from h5i itself instead of stopwatching.
5. Write up `docs/benchmarks/env-overhead.md`: a table of median wall-clock per (tier, workload), the delta vs bare, plus host info (OS, kernel, WSL2 or bare metal, Landlock ABI from `h5i env probe`, podman version if container ran).

## Acceptance criteria

- [ ] `docs/benchmarks/env-overhead.md` exists with median-of-5 timings for at least the `workspace` and one kernel tier (`process` or `supervised`), and `container` if your host has rootless Podman.
- [ ] Both a short and a long workload are measured, so setup cost and steady-state overhead are distinguishable.
- [ ] The doc records the host capability report so numbers from different machines can be compared.
- [ ] A "how to reproduce" section with the exact commands, so the benchmark can be re-run per release.

## Hints

- `h5i env create` fails closed with a clear message if a tier cannot actually enforce on your host (e.g. AppArmor-restricted userns). Report that as "unsupported on this host" in the table rather than fighting it.
- Memory limits use RLIMIT_DATA on kernel tiers, so an artificially tiny `mem` limit can OOM a big build. Use the default profile limits for the benchmark.
- Clean up with `h5i env rm bench-<tier>` when done.

floze-the-genius:
## Summary

- publish median-of-five wall-clock results for bare, `workspace`, and `process` execution
- record the Ubuntu/OrbStack host capability report, raw samples, and unsupported tiers
- include a copy-paste reproduction harness and explain the 25 ms workspace polling effect

## Validation

- `H5I_SKIP_WEB_BUILD=1 cargo build --release --no-default-features`
- `h5i env probe`
- `h5i env capabilities --json`
- one warm-up plus five measured runs per workload/tier, with rotating execution order
- `git diff --check`

`container` was not benchmarked because rootless Podman was unavailable on the temporary Ubuntu 24.04 VM.

Closes #353

### DIFF
--- a/docs/benchmarks/env-overhead.md
+++ b/docs/benchmarks/env-overhead.md
@@ -1,0 +1,182 @@
+# `h5i env run` isolation overhead
+
+`h5i env run` adds little steady-state overhead on this host. The `process`
+tier was within 1% of bare execution for a 338 ms CPU workload. Short
+`workspace` runs exposed a separate cost: the current completion loop polls at
+25 ms intervals, so a 4 ms command was reported as 30 ms at the median.
+
+These results are one Apple Silicon VM sample, not a cross-platform claim.
+
+## Results
+
+Each cell is the median wall-clock time from five measured runs after one
+warm-up. Bare commands were timed around `subprocess.run()` with
+`time.perf_counter_ns()`. Sandboxed times came from the `wall` field printed by
+`h5i env run` and stored in the env exec event.
+
+| Workload | Isolation | Median wall | Delta vs bare |
+|---|---:|---:|---:|
+| Python startup | bare | 4.4 ms | baseline |
+| Python startup | `workspace` | 30.0 ms | +25.6 ms (+579.1%) |
+| Python startup | `process` | 5.0 ms | +0.6 ms (+13.2%) |
+| SHA-256 loop | bare | 338.0 ms | baseline |
+| SHA-256 loop | `workspace` | 357.0 ms | +19.0 ms (+5.6%) |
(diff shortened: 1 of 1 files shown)


Koukyosyumei:
LGTM. Thank you so much!

PR #365, opened 2026-07-21
feat(msg): add JSON team roster output

### LINKED ISSUE #337: Add `--json` to `h5i msg team`
## Summary

`h5i msg inbox` and `h5i msg history` gained `--json` in #324, but `h5i msg team` (the roster of known agents on the channel) still only offers `--plain`. Orchestration scripts that want to know "which agents exist on this clone and when were they last seen" have to parse the styled table.

## Proposed behavior

```bash
h5i msg team --json
```

emits a JSON array like:

```json
[
  {"name": "claude", "last_seen": "2026-07-12T19:58:17Z", "you": true},
  {"name": "codex",  "last_seen": "2026-07-11T08:02:00Z", "you": false}
]
```

An empty roster emits `[]`.

## Where to look

- Handler: `src/cli/msg.rs`, the `Some(MsgCommands::Team)` arm around line 637. It already has everything: `msg::team(git)` returns `Vec<(name, last_seen_ts)>` and `msg::read_identity(&h5i_root)` gives the "you" marker.
- Core: `crates/h5i-core/src/msg.rs`, `pub fn team` at line 766 (no changes needed there).
- Clap: add the flag to the `Team` variant (line 217 of `src/cli/msg.rs`), next to the existing `--plain` pattern used by the other msg subcommands.
- Precedent: the `--json` help text on `msg history` notes that fields are untrusted collaborator input. Agent names in the roster come from pulled data too, so reuse the same wording in the flag's doc comment.

## Acceptance criteria

- `h5i msg team --json` prints valid JSON only (no ANSI codes), `[]` when the roster is empty.
- `--plain` and default styled output are unchanged.
- A unit or integration test asserting the JSON shape for a roster with at least two agents, one being the caller.

AnayGarodia:
Fixes #337.

Adds `h5i msg team --json` with a stable array of `{name, last_seen, you}` records. JSON mode emits only valid JSON, including `[]` for a roster with no messages; the default styled roster and global `--plain` behavior are unchanged.

Agent names and timestamps are serialized directly as JSON values rather than rendered, so pulled collaborator data cannot inject ANSI output. The manual now documents the machine-readable form.

Verification:
- `cargo test --test msg_integration` — 41 passed
- `git diff --check`

Committed with `h5i capture commit --agent codex` and linked context/test provenance.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -2860,6 +2860,7 @@ h5i msg history --with codex     # full conversation log
 h5i msg history --json           # machine-readable message history
 h5i msg replay --with codex      # replay the log as a live feed (1s between messages)
 h5i msg team                     # known agents
+h5i msg team --json              # machine-readable roster with last-seen and you fields
 ```
 
 Add `--plain` to any read command for greppable, uncoloured output.
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -3528,6 +3528,7 @@ h5i msg history --with codex     # full conversation log
 h5i msg history --json           # machine-readable message history
 h5i msg replay --with codex      # replay the log as a live feed (1s between messages)
 h5i msg team                     # known agents
+h5i msg team --json              # machine-readable roster with last-seen and you fields
 </code></pre>
 <p>Add <code>--plain</code> to any read command for greppable, uncoloured output.</p>
 <h3 id="delivery-modes">Delivery modes</h3>
--- a/man/man1/h5i.1
+++ b/man/man1/h5i.1
@@ -2147,10 +2147,13 @@ List the known agents on this repo's message roster
 .ie \n(.g .ds Aq \(aq
 .el .ds Aq '
 .SS SYNOPSIS
-\fBteam\fR [\fB\-h\fR|\fB\-\-help\fR] 
+\fBteam\fR [\fB\-\-json\fR] [\fB\-h\fR|\fB\-\-help\fR] 
 .ie \n(.g .ds Aq \(aq
 .el .ds Aq '
 .SS OPTIONS
+.TP
+\fB\-\-json\fR
+Emit roster records as JSON. Treat fields as untrusted collaborator input
 .TP
 \fB\-h\fR, \fB\-\-help\fR
 Print help
--- a/src/cli/msg.rs
+++ b/src/cli/msg.rs
@@ -214,7 +214,11 @@ pub enum MsgCommands {
     },
 
     /// List the known agents on this repo's message roster.
-    Team,
+    Team {
+        /// Emit roster records as JSON. Treat fields as untrusted collaborator input.
+        #[arg(long)]
+        json: bool,
+    },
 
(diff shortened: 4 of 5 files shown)


Koukyosyumei (changes requested):
Thanks for this — the shape matches #337 exactly, and I appreciate that `--json` short-circuits before the styled path so the empty roster is `[]` rather than the "no agents yet" line. One requested change, on the `you` field.

## What I verified locally (trial merge with current `main`; the branch is 7 commits behind, merge is clean)

- `cargo test --test msg_integration` : 41 passed.
- `./scripts/gen_man.sh` and `python3 scripts/gen_manual.py` on the merged tree regenerate identically, so `manuals fresh` stays green even though `main` has since bumped to 0.3.0.
- Smoke: valid JSON only, no ANSI, `[]` on an empty roster, styled and `--plain` output unchanged.

## Requested change: `you` uses the untrustworthy identity slot

`msg::read_identity` returns the raw stored slot, which `send_msg` itself declines to trust in a shared clone (`crates/h5i-core/src/msg.rs:351-358`):

> Persist the sender as the local default **only in a solo clone**. In a shared clone this single slot is ambiguous — `resolve_identity` refuses to trust it — so writing it just churns shared state and **risks misleading a tool that reads the file directly**.

`msg team --json` is now that tool. In the shared-clone model h5i documents (several agents, one clone, identity injected as `$H5I_AGENT`), the slot holds whoever sent last, so `you` lands on the wrong agent:

```console
$ H5I_AGENT=alice h5i msg team --json
[
  { "name": "alice", "last_seen": "...", "you": false },
  { "name": "bob",   "last_seen": "...", "you": true }    # bob sent last
]
```

The sibling `--json` command already does the right thing — `msg inbox` resolves via `msg::resolve_identity` (`src/cli/msg.rs:529`). Same one-liner here:

```rust
let me = msg::resolve_identity(&h5i_root, None).ok();
```

`explicit > $H5I_AGENT > stored`, and in an ambiguous shared clone it returns `Err`, so `you` is simply `false` everywhere rather than wrong. The roundtrip test then needs the env-injected identity instead of the stored one; the harness already has a helper (`h5i_as`, `tests/msg_integration.rs:75`):

```rust
let team_json = out_str(&a.h5i_as("alice", &["msg", "team", "--json"]));
```

I applied both locally and re-ran `cargo test --test msg_integration` : 41 passed.

The styled arm (`src/cli/msg.rs:664`) mislabels `(you)` the same way. That is pre-existing and not something this PR introduced, but it is the same one-line change and it would keep the two views in agreement. Happy either way — fixing it here is welcome, leaving it for a follow-up is fine.

## Optional

`json!` gives alphabetical key order (`last_seen`, `name`, `you`) rather than the order in the issue example. Purely cosmetic for a consumer, no action needed unless you would rather use a small `#[derive(Serialize)]` struct.

Happy to approve once `you` resolves through `resolve_identity`. Thanks also for the `[]`-before-any-messages test — that is the case scripts trip over first.

Koukyosyumei:

PR #366, opened 2026-07-21
feat(filters): add Playwright output filter

### LINKED ISSUE #202: Add an output filter for `playwright`
## Add an output filter for `playwright`

h5i compresses noisy command output before it reaches an agent's context (this powers `h5i capture run`). Filters are small **declarative TOML rule files** under [`assets/filters/`](https://github.com/h5i-dev/h5i/tree/main/assets/filters), embedded into the binary at build time. This is a self-contained good-first-issue: **add a filter for `playwright`**. No changes to core logic are needed.

### Why
Playwright's reporter prints per-test lines and a long trace; the signal is failing tests, the error/expected-vs-received, and the pass/fail counts.

### What to keep / strip
- **Keep (the signal):** failing test titles, `Error:`/`expect(...)` diffs, and the `N passed / M failed` summary.
- **Strip (the noise):** passing-test lines, worker setup chatter, and blank lines.

### Steps
1. Create `assets/filters/playwright.toml`.
2. Match the command with a `match_command` regex, e.g. `match_command = "\\bplaywright\\b"`. Make it match typical invocations but not unrelated commands.
4. Add **2–3 inline golden tests** (`[[tests.playwright]]`): a normal run, an error/failure run, and a clean/empty run if relevant.

### Golden-test shape
```toml
[filters.playwright]
description = "…"
match_command = "\\bplaywright\\b"
strip_ansi = true

[[tests.playwright]]
input = """…raw output…"""
```

### How to verify
```bash
```

### Reference
The upstream [`rtk`](https://github.com/rtk-ai/rtk) project has a `playwright` adapter — see its `src/filters/` and `src/cmds/` for inspiration on what matters in the output. h5i's filters are derived from rtk (see `assets/filters/NOTICE`).

### Acceptance criteria
- [ ] `assets/filters/playwright.toml` added with a clear `description`
- [ ] `match_command` matches typical `playwright` invocations and not unrelated ones

AnayGarodia:
Fixes #202.

Adds an embedded declarative Playwright filter that removes per-test progress, passing-test rows, blank lines, worker-count chatter, and report-server hints while preserving failing titles, `Error:` blocks, expected/received diffs, source excerpts, and pass/fail totals.

The command regex covers direct, npx, bunx, yarn, pnpm, and pnpm-exec invocations while rejecting similar executable names such as `playwright-helper`.

Verification:
- Four inline golden cases: sizeable all-pass, assertion failure, setup error, and empty output
- `cargo test -p h5i-core builtin_golden_tests_pass`
- `cargo test -p h5i-core --test filter_quality` — 15 passed, including a >1,000-token fixture constrained below 10% of raw tokens
- `git diff --check`

Committed with `h5i capture commit --agent codex` and linked test/context provenance.

### DIFF
--- a/assets/filters/playwright.toml
+++ b/assets/filters/playwright.toml
@@ -1,0 +1,106 @@
+[filters.playwright]
+description = "Compact Playwright test output — keep failures, assertion diffs, and summary counts while dropping passing progress noise"
+match_command = "^(?:(?:npx|bunx|yarn)\\s+|pnpm\\s+(?:exec\\s+)?)?playwright(?:\\s|$)"
+strip_ansi = true
+strip_lines_matching = [
+  "^\\s*$",
+  "^Running \\d+ tests? using \\d+ workers?",
+  "^\\s*\\[\\d+/\\d+\\]\\s+",
+  "^\\s*[✓✔]\\s+",
+  "^\\s*To open last HTML report run:",
+  "^\\s*Serving HTML report at ",
+]
+truncate_lines_at = 240
+max_lines = 100
+on_empty = "playwright: no output"
+
+[[tests.playwright]]
+name = "all-pass run keeps only the summary"
+input = """
+Running 12 tests using 3 workers
+
+[1/12] [chromium] › tests/auth.spec.ts:4:1 › signs in
+  ✓  1 [chromium] › tests/auth.spec.ts:4:1 › signs in (421ms)
+[2/12] [chromium] › tests/auth.spec.ts:12:1 › signs out
+  ✓  2 [chromium] › tests/auth.spec.ts:12:1 › signs out (308ms)
+[3/12] [chromium] › tests/cart.spec.ts:5:1 › adds an item
+  ✓  3 [chromium] › tests/cart.spec.ts:5:1 › adds an item (512ms)
+[4/12] [chromium] › tests/cart.spec.ts:16:1 › updates quantity
+  ✓  4 [chromium] › tests/cart.spec.ts:16:1 › updates quantity (442ms)
+[5/12] [chromium] › tests/cart.spec.ts:27:1 › removes an item
+  ✓  5 [chromium] › tests/cart.spec.ts:27:1 › removes an item (391ms)
+[6/12] [firefox] › tests/auth.spec.ts:4:1 › signs in
+  ✓  6 [firefox] › tests/auth.spec.ts:4:1 › signs in (476ms)
+[7/12] [firefox] › tests/auth.spec.ts:12:1 › signs out
+  ✓  7 [firefox] › tests/auth.spec.ts:12:1 › signs out (355ms)
+[8/12] [firefox] › tests/cart.spec.ts:5:1 › adds an item
+  ✓  8 [firefox] › tests/cart.spec.ts:5:1 › adds an item (548ms)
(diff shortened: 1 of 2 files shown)


Koukyosyumei (changes requested):
Nice work — the four golden cases are well chosen (the setup-error one especially, since that path has no test titles at all), and rejecting `playwright-helper` while still routing `npx`/`bunx`/`yarn`/`pnpm exec` is a real improvement over the `\bplaywright\b` sketch in #202, which would have matched the helper. One requested change, on the line cap.

## What I verified locally (trial merge with current `main`; the branch is 7 commits behind, merge is clean)

- `cargo test -p h5i-core --lib filter_rules` (includes `builtin_golden_tests_pass` and `registry_is_name_sorted_and_unambiguous`) : 9 passed.
- `cargo test -p h5i-core --test filter_quality` : 15 passed.
- Routing spot-checks beyond the ones in your test: `./node_modules/.bin/playwright test` also routes correctly, since `command_string` basenames `argv[0]`.

## Requested change: the summary counts are lost on a failure-heavy run

`max_lines` truncates from the **head** (`crates/h5i-core/src/filter_rules.rs:324-330`), so once the surviving lines exceed 100 it is the tail that goes — and in Playwright the tail is the `N failed` / `N passed` block that #202 lists as signal to keep. Playwright reprints a full error, code excerpt, `at` frame and attachment lines per failure, so 100 lines is roughly 8 failures.

Repro with a synthetic 40-test run, 12 failures, 277 raw lines:

```
$ h5i capture run -- playwright test | tail -3
        at /app/tests/checkout_9.spec.ts:23:5
... (46 lines truncated)
```

Failures 10 to 12 are gone, and so are `12 failed` and `28 passed (41.2s)` — an agent reading that summary cannot tell how bad the run was.

The engine's head+tail pair fixes it. Adding these two lines above the existing `max_lines`:

```toml
head_lines = 70
tail_lines = 25
max_lines = 100
```

gives, on the same fixture, `12 failed` and `28 passed (41.2s)` both present, with an explicit `... (N lines omitted)` marker in the middle, and all four inline golden tests still pass (they are well under the caps, so their `expected` blocks do not change). Note that head 80 / tail 20 does *not* work: 80 + marker + 20 = 101 exceeds `max_lines`, and the one line that gets cut is the totals. Precedent for the combination is `trunk-build.toml`.

## Notes, no action needed

- `npm exec playwright test` routes to the `npm` rule, not this one: rules are first-match in name order and `npm` sorts before `playwright`. Nothing this PR can do about it, and the degradation is graceful (npm's filter keeps the failures, just compresses less). Recording it here so it is not mistaken for a gap in your regex.
- `jest.toml` and `vitest.toml` have the same head-truncation weakness. That is on me, not on you — I will open a follow-up.

## Nit

The new test code is laid out more loosely than the rest of `filter_quality.rs`, which keeps these on one line:

```rust
let res = filter(&raw, &cfg(OutputKind::Auto, Some(argv(&["npx", "playwright", "test"]))));
```

The file is not `rustfmt`-enforced (see the note in `.github/workflows/test.yaml`), so this is style consistency only.

Happy to approve as soon as the totals survive a failure-heavy run. Thanks for the `>1,000-token` fixture in `filter_quality.rs` — that is the assertion that keeps this rule honest.

Koukyosyumei:

PR #379, opened 2026-07-23
Add JSON output to recall search

### LINKED ISSUE #336: Add `--json` to `h5i recall search` for parity with `recall objects`
## Summary

`h5i recall objects --json` exists precisely so headless consumers get a typed feed instead of regex-parsing text. Its deeper sibling `h5i recall search` (which queries the normalized findings inside captures) still has no `--json`, so the one command designed for programmatic questions like "has this fingerprint failed before?" can only answer in human-formatted text.

## Proposed behavior

```bash
h5i recall search "connection refused" --severity error --json
h5i recall search --fingerprint 3fa9 --json
```

emits a JSON array of hits. Each hit should carry the capture metadata (object id, tool, cmd, exit/status, branch, env_id, timestamp) plus the matched finding fields (message, rule, path/location, severity, kind, fingerprint), mirroring what the text view shows.

## Where to look

- Handler: `src/cli/objects.rs`, the search arm around line 999 where `objects::search_manifests(&newest, &filters)` is called and the hits are printed.
- Core: `crates/h5i-core/src/objects.rs`, `pub fn search_manifests` (around line 1944) returns `SearchHit` values; serializing those (plus the parent manifest fields) is the bulk of the work.
- Precedent to copy: the `--json` arm of `recall objects` list in the same file, which already defines the typed-feed shape and field-naming conventions. Keep the two consistent where fields overlap.

## Acceptance criteria

- `h5i recall search <query> --json` prints valid JSON only, respecting every existing filter flag (`--severity`, `--rule`, `--path`, `--fingerprint`, `--tool`, `--env`, `--since`, `--limit`).
- Zero matches emits `[]`, not a human "no results" line.
- Human output without the flag is unchanged.
- An integration test that captures a failing command via `h5i capture run`, then asserts `recall search --json` finds it and the JSON parses with the expected fields.

snowyukitty:
Adds `--json` to `h5i recall search`, returning the same capture metadata as `recall objects --json` plus only the findings matched by the query and filters. Empty searches return `[]`; the existing human output is unchanged.

Adds integration coverage for typed metadata, finding fields, and zero-match output, and documents the new flag.

Closes #336

### Testing

- `cargo test --locked --test objects_e2e search_json_emits_typed_hits_and_an_empty_array -- --exact --nocapture`
- `cargo test --locked --test objects_e2e search_ -- --nocapture`
- `cargo test --locked -p h5i-core objects::tests::search -- --nocapture`
- `cargo build --verbose --all-targets`
- `cargo check --locked --target x86_64-pc-windows-msvc`
- `cargo test --verbose` (217 passed; the same seven WSL process-tier integration failures reproduce on the exact base)

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -47,7 +47,7 @@ emits `--json` is safe to consume from a script or another agent.)
 | Record what an AI change was (model, agent, prompt) | [`h5i capture commit`](#h5i-capture-commit) | Provenance stored in Git notes; the human prompt is auto-captured in Claude Code. |
 | Keep noisy command output out of my context | [`h5i capture run -- <cmd>`](#h5i-capture-run) | Prints a compact summary, stores the full output out-of-band, passes the exit code through. |
 | Read history with AI authorship | [`h5i recall log`](#h5i-recall-log), [`recall blame`](#h5i-recall-blame) | Who / which model wrote each line, with the prompt and test metrics. |
-| Inspect or grade what commands actually ran | [`h5i recall objects [--json]`](#h5i-recall-object--objects), [`recall search`](#h5i-recall-object--objects) | Typed feed: `cmd`, `exit`, **action** (test/build/read/write/egress), **tool**, validated **status**, enforced **egress** verdicts. |
+| Inspect or grade what commands actually ran | [`h5i recall objects [--json]`](#h5i-recall-object--objects), [`recall search [--json]`](#h5i-recall-object--objects) | Typed feed: `cmd`, `exit`, **action** (test/build/read/write/egress), **tool**, validated **status**, enforced **egress** verdicts. |
 | Track my goal + reasoning across a task | [`h5i recall context`](#h5i-recall-context) | Goal + auto-derived reasoning trace; snapshotted per commit. |
 | Triage the risk of an AI branch | [`h5i audit review`](#h5i-audit-review), [`audit scan`](#h5i-audit-scan) | Rank commits by risk; scan reasoning traces for prompt-injection. |
 | Score how well the engineer prompted | [`h5i audit maturity [--json]`](#h5i-audit-maturity) | Headless / CI prompt-maturity score (feeds merge confidence). |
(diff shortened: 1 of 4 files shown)


Koukyosyumei (approved):
Approving. This is a clean, well-scoped addition: the JSON path is additive, the human output is untouched, and the field set is an exact match for `recall objects --json`, which is what makes the two feeds interchangeable for a headless consumer.

## What I verified locally (worktree at `dfce771`, merged trial against `main`)

- `cargo test --test objects_e2e search_` : 8 passed, including the new `search_json_emits_typed_hits_and_an_empty_array`.
- `python3 scripts/gen_manual.py` : regenerates `docs/manual/index.html` byte-identically, so the committed HTML is genuinely fresh and the `manuals fresh (man + html)` gate will pass.
- `./scripts/gen_man.sh` : `man/man1/h5i.1` unchanged, as expected. `recall` is a `[REST]` passthrough in the clap tree, so verb-level flags never reach the man page. No man regen was needed here.
- Trial merge with current `main` : clean.

## Design notes

- Field parity with `ObjJson` is exact (14 fields, same names, same `skip_serializing_if`), and `.take(limit)` matches the human path, so `--limit` behaves identically in both modes. Good.
- Emitting the full matched finding set per capture (rather than the human path's `PER_CAPTURE = 8` cap) is the right call for a machine feed: a consumer that wants every hit should not have to fall back to `recall object <id>`.

## Non-blocking suggestions

1. **Drift risk in the duplicated struct.** `SearchJson` restates every `ObjJson` field. The next field added to `recall objects --json` will silently not appear in `recall search --json`, and nothing in the test suite would catch it. Consider one shared struct with an optional findings slot:

   ```rust
   #[serde(skip_serializing_if = "Option::is_none")]
   findings: Option<Vec<&'a h5i_core::structured::Finding>>,
   ```

   plus a small `fn row(m: &Manifest, store: &LocalStore, findings: Option<...>) -> ObjJson` used by both arms. That also removes the third copy of the `serde_json::to_value(status)` round-trip, which already exists as the private `objects::enum_str` in `h5i-core`; making that `pub` would let both call sites share it.

2. **Silent truncation.** When `hits.len() > limit` the human path prints `— showing {limit}`, but the JSON array just ends, so a consumer cannot distinguish 20 results from 20-plus. This matches `recall objects --json` exactly, so I would not change it in this PR; worth considering a `truncated` marker for both feeds later if CI graders start paging.

Nothing here blocks the merge. Thanks for including the zero-match `[]` case and the `MANUAL.md` capability-table row, both of which are easy to forget.

PR #382, opened 2026-07-26
feat(env): add JSON allowlist output

### LINKED ISSUE #339: Add `--json` to the `h5i env allow` list view

lntutor:
Summary:
- add list-only --json output to h5i env allow
- include the persisted allowlist path and normalized rules
- reject --json with a rule or --remove

Fixes #339

Validation:
- cargo test --test env_integration env_allow_add_list_remove_and_in_box_refusal -- --exact

Note: repository-wide cargo fmt --check currently reports formatting drift outside this change; the focused test and diff whitespace checks pass.


Koukyosyumei (changes requested):
The behavior here is right and the guard rails are the ones I would have asked for. Two small follow-ups before merge, both mechanical.

## What I verified locally (worktree at `9fca2c0`, and again merged against current `main`)

- `cargo test --test env_integration env_allow` : `env_allow_add_list_remove_and_in_box_refusal` and `process_tier_env_allowlist` both pass, on the PR head and on a trial merge with `main` (this branch is 16 commits behind, merge is clean).
- `./scripts/gen_man.sh` on the **merged** tree : `man/man1/h5i.1` regenerates identically, so the `manuals fresh` gate stays green even though `main` has since touched the man page and bumped to 0.3.0.
- Manual smoke: `env allow --json` on an empty allowlist emits `{"path": "<would-be path>", "rules": []}` with exit 0, and `--json` with `--remove` or a rule exits 1 with the intended message.

Reporting the path even before the file exists, and keeping `path` as `null` rather than omitting the key, are both the right choices for a headless consumer. Confining `--json` to list mode instead of inventing a result object for the mutating paths is also the correct scope call.

## Requested changes

1. **`MANUAL.md` still advertises the old signature.** The env reference row reads:

   ```
   | `h5i env allow [<host>] [--remove]` | Manage the persistent user-level egress allowlist ...
   ```

   The man page is generated, but `MANUAL.md` is hand-maintained and is the narrative reference the man page's SEE ALSO points at, so it now understates the surface. Please update that row to `[--remove] [--json]` with a clause for the list-only JSON output, then regenerate the rendered page:

   ```bash
   python3 -m pip install -r scripts/requirements.txt && python3 scripts/gen_manual.py
   git add MANUAL.md docs/manual/index.html
   ```

   Worth considering the "What h5i can do" table too: the `env probe` / `env capabilities --json` row is where someone hunting for machine-readable egress state would look.

2. **The new `match rule {` leaves its arms under-indented.** In `src/cli/env.rs` the arms sit at the same column as the `match` itself, and the two closing braces stack:

   ```rust
   match rule {
   None => {
       ...
   }
   Some(raw) => {
   ```

   You are right that the file is not `rustfmt`-clean today (the whole `pub fn run` body is over-indented by one level because of the bare wrapper block, which is pre-existing and out of scope). But this block is now inconsistent with its own surroundings, which makes the diff harder to read than it needs to be. Please indent the arms one level under the new `match`.

## Optional

Instead of the manual guard, clap can express the exclusion declaratively:

```rust
#[arg(long, conflicts_with_all = ["rule", "remove"])]
json: bool,
```

That removes the `bail!` and the extra nesting entirely, and the conflict then shows up in `--help` and the man page. The tradeoff is that clap's generic conflict message is less instructive than your "can only be used when listing the allowlist", and it exits 2 rather than 1, so the assertion in `env_integration.rs` would need adjusting. Your call. I am fine with the explicit `bail!` as written.

Happy to approve as soon as the manual row and the indentation are in. Thanks for the in-box refusal check and the two negative-path assertions in the test.

lntutor:
Addressed the requested follow-ups in 7684e90: `MANUAL.md` and the regenerated manual page now document `h5i env allow [<host>] [--remove] [--json]` and its list-only headless output; the new `match rule` arms are consistently indented. Local `cargo check` and `cargo test --test env_integration env_allow` pass, and the fresh CI matrix is green. Please re-review when convenient.

lntutor:
Following up on the current head `7684e90189f27fffa44f3e48979a425fb07defdb`: the requested `MANUAL.md`/rendered manual update and the `match rule` indentation cleanup are both on this SHA, and the full Rust matrix is green (`build · test · clippy · fmt`, `manuals fresh`, and all cross-checks). This should be ready for re-review when convenient.

lntutor:
@Koukyosyumei Current head `7684e90189f27fffa44f3e48979a425fb07defdb` is still the branch with the `MANUAL.md`/rendered manual update and the `match rule` indentation cleanup from your July 30 review, and the full Rust/manual matrix remains green on this SHA. When convenient, could you please take another look?

Koukyosyumei (approved):
Both follow-ups are in. Verified on a trial merge with current `main`: `env_allow` tests pass, clippy clean, and both `gen_man.sh` and `gen_manual.py` regenerate identically. Approving — thanks for the patient follow-up.

PR #417, opened 2026-08-07
fix: refuse Intel macOS in installer

### LINKED ISSUE #413: install.sh silently installs an arm64 binary on Intel Macs
## What happens

`install.sh` maps both macOS architectures to the same release target:

```sh
case "${os}-${arch}" in
  linux-x86_64)  target="x86_64-unknown-linux-musl" ;;
  linux-aarch64) target="aarch64-unknown-linux-musl" ;;
  macos-x86_64 | macos-aarch64) target="aarch64-apple-darwin" ;;
esac
```

`.github/workflows/release.yaml` only builds four targets, and the only macOS
one is `aarch64-apple-darwin`. So on an Intel Mac the script downloads an arm64
binary, installs it to `/usr/local/bin/h5i`, prints the success line, and the
first `h5i --help` fails with a bad CPU type error. The installer reports
success for something that cannot run.

Second, smaller problem in the same block: there is no `*)` arm. If `os`/`arch`
detection ever produces a pair not listed, `target` stays empty, the archive
name becomes `h5i-v0.3.1-.tar.gz`, and the user gets a raw curl 404 instead of a
sentence explaining the platform is unsupported.

## What it should do

Fail before downloading, with a message that says what is actually available:

* `macos-x86_64`: exit with a note that only Apple Silicon builds are published
  today, and that Rosetta 2 is not a workaround for a native binary. Mention
  `cargo install --path .` as the source route.
* unknown pair: exit with `Unsupported platform: ${os}-${arch}` rather than
  falling through with an empty `target`.

Keep `macos-aarch64` mapping to `aarch64-apple-darwin` as it does now.

## Where

* `install.sh`, the "map to release target triple" block (around line 30).
* Target list to check against: `.github/workflows/release.yaml`, the build
  matrix around line 72.

## How to verify

The script reads the arch from `uname -m`, so you can exercise both paths
without an Intel Mac by stubbing the detection locally, or by extracting the
case block into a small shell function and calling it with each pair. Confirm
the Apple Silicon and both Linux paths still resolve to the same triples they do
today.

## Why this is a good first issue

Self-contained: one `case` block in one POSIX shell script, no Rust, no test
harness to learn. The behavior it fixes is a silent wrong answer, which is the
kind worth removing.

slegarraga:
Closes #413

## Summary

Fails before download on unsupported platform pairs instead of installing a binary that cannot run.

- `macos-aarch64` still maps to `aarch64-apple-darwin`.
- `macos-x86_64` exits with a clear message that only Apple Silicon builds are published, notes Rosetta is not a workaround, and points to `cargo install --path .`.
- Any unknown `os-arch` pair exits with `Unsupported platform: <os>-<arch>`.

## Verification

- `sh -n install.sh` passes.
- `git diff --check` passes.
- Supported Linux and Apple Silicon triples are unchanged.

Signed off with DCO.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -276,7 +276,12 @@ jobs:
             |--------|---------|
             | `x86_64-unknown-linux-musl` | Linux x86_64 (static, glibc-free) |
             | `aarch64-unknown-linux-musl` | Linux arm64 (static, glibc-free) |
-            | `aarch64-apple-darwin` | macOS (Apple Silicon + Rosetta) |
+            | `aarch64-apple-darwin` | macOS Apple Silicon (arm64) |
             | `x86_64-pc-windows-msvc` | Windows x86_64 |
+
+            Intel Macs have no published build: Rosetta 2 translates x86_64 to
+            arm64, not the reverse, so the `aarch64-apple-darwin` archive will
+            not run there. Build from source instead:
+            `cargo install --git https://github.com/${{ github.repository }}`
 
             Each archive is accompanied by a `.sha256` checksum file.
--- a/docs/install.sh
+++ b/docs/install.sh
@@ -34,7 +34,23 @@ main() {
   case "${os}-${arch}" in
     linux-x86_64)  target="x86_64-unknown-linux-musl" ;;
     linux-aarch64) target="aarch64-unknown-linux-musl" ;;
-    macos-x86_64 | macos-aarch64) target="aarch64-apple-darwin" ;;
+    macos-aarch64) target="aarch64-apple-darwin" ;;
+    # Rosetta 2 translates x86_64 to arm64, not the reverse, so the published
+    # Apple Silicon archive cannot run here. Fail before the download rather
+    # than install a binary that will not execute.
+    macos-x86_64)
+      echo "Unsupported platform: macos-x86_64." >&2
+      echo "Only Apple Silicon macOS builds are published; Rosetta 2 cannot run a native arm64 binary on an Intel Mac." >&2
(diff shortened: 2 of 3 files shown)


Koukyosyumei (approved):
Approved. Thanks for this. The Intel Mac case is a real footgun and the fix is the right shape: fail before the download rather than install a binary that cannot execute.

I rebased the branch onto current `main` and force-pushed, because it had gone stale in three ways since you opened it. Flagging each so nothing is a surprise:

1. **`install.sh` was rewritten on `main`.** Everything now lives in a `main()` function called on the last line, so a truncated `curl … | sh` cannot execute a half-downloaded prefix. Your branch was based on the pre-rewrite file, so merging as-is would have reverted that guard. The platform arms are now applied inside the wrapper.

2. **The `compiled_features()` commit was dropped** — it already landed as #416, and the merged version is broader than the one here: it also reports `share` and `share-tunnel`, which this branch's variant would have omitted from `--version --json`.

3. **`docs/install.sh` needed the same change.** That copy is what `h5i.dev/install.sh` serves, and CI now compares the two byte for byte. It is squashed into your commit.

One content change beyond the rebase: the error message pointed at `cargo install --path .`, which nobody who arrived via `curl | sh` can run — they have no checkout. It now says `cargo install --git https://github.com/h5i-dev/h5i`.

Also added a second commit fixing the release notes, which described `aarch64-apple-darwin` as "Apple Silicon + Rosetta". That has the translation direction backwards and is the same misconception behind #413, so it seemed worth correcting in the same pass.

Your authorship and sign-off are preserved on the installer commit.

Issue #444, opened 2026-08-07
A microVM outlives an abnormal h5i exit


Koukyosyumei:
`SandboxGuard` reaps the named `msb` sandbox only through `Drop`.

crates/h5i-sandbox/src/microvm.rs:730-760.

On SIGTERM, SIGKILL, or in a panic-abort build, the named sandbox persists. The module itself notes that `msb` keeps *named* sandboxes around and that h5i "takes responsibility for reaping it" — which is precisely what does not happen on these paths.

The container tier has `--rm` as a backstop; this tier has none.

**Fix.** Install a signal handler that reaps, and/or sweep orphaned `h5i-*` sandboxes at startup.

Koukyosyumei:
Fixed in #465.

Issue #437, opened 2026-08-07
Egress proxy spawns an unbounded thread per connection; the tee shim writes uncapped output to the host spool


Koukyosyumei:
Two host-side resources are driven directly from inside the box with no limit.

**Threads.** crates/h5i-sandbox/src/container.rs:478-500 spawns one thread per accepted connection, plus a second in `splice`, with no cap and no backpressure. A box can open thousands of proxy connections and exhaust host threads.

**Disk.** The tee shim (container.rs:784-788) writes the full stdout and stderr of every observed command into the host-side spool with no size limit:

```sh
tee "$b.out" < "$b.po" &
tee "$b.err" < "$b.pe" >&2 &
```

The ingest side caps what it *reads*, but the box controls what it *writes*, so `yes > /dev/stdout` fills the operator's disk.

**Fix.** Bound the proxy with a small worker pool, and have the shim cap each capture file (truncate with a marker) rather than relying on ingest-side caps.

Koukyosyumei:
Fixed in #465.

Issue #435, opened 2026-08-07
Pre-auth slowloris exhausts the auth proxy and costs the box its credentialed egress


Koukyosyumei:
`read_head` reads one byte per `read()` with only a per-read 60 s timeout, and the shared-secret gate runs *after* the whole head is read, so an unauthenticated client can hold a concurrency slot indefinitely.

crates/h5i-sandbox/src/auth_proxy.rs:540-554 (`read_head`), :588-589, :349-354 (`InFlightSlot`), gate at :595.

`client.set_read_timeout(Some(60s))` bounds a single `read`, not the head. The loop only exits on `\r\n\r\n` or 64 KiB. There is no overall deadline and no per-peer limit.

**Repro.** Any local process — the exact threat actor the token gate exists to stop, per the module doc at auth_proxy.rs:29-34 — opens 64 connections and dribbles one header byte every 59 seconds. Each occupies an `InFlightSlot`. The 65th connection, the agent's actual API call, gets `503 Service Unavailable` at :349-353. The agent loses model access for the life of the box and cannot fall back, because its only credential is the dummy.

**Related.** `MAX_REQUEST_BODY` (32 MiB) is enforced per request while `MAX_IN_FLIGHT` is 64, so a box looping requests with `Content-Length: 33554432` can pin 2 GiB of host allocation (`vec![0u8; req.content_length]`, :612-622) before any body byte arrives. The comment at :156-160 reasons about a single request only; there is no aggregate byte budget.

**Fix.** Apply an overall head deadline, move the token gate before the full head read where possible, add a per-peer connection cap, and add an aggregate body budget.

Koukyosyumei:
Fixed in #465.

Issue #432, opened 2026-08-07
The web forward hangs and stops serving when the box's stream closes first


Koukyosyumei:
`Forward::handle` only handles one shutdown ordering.

crates/h5i-core/src/view.rs:859-866. The `out` thread's `io::copy(upstream → client_out)` returns on upstream EOF, but nothing shuts down the client socket. The main thread stays blocked in `pump_input(&mut client, ...)` reading a client that will never close, because the page never learns the stream ended and keeps the socket open. Only the reverse ordering is handled: after `pump_input` returns, `upstream_out.shutdown(...)` unblocks the copy thread.

**Repro.** Open a viewer tab, then let the box browser exit or run `agent-browser stream disable`.

Because `serve()` is one-connection-at-a-time (view.rs:813-822), the forward then accepts nothing further and `record_session` is never written until the human happens to close the tab.

**Related, same file.** `read_head` (view.rs:1008-1022) has no timeout, so a browser speculative preconnect that opens the port and sends nothing stalls the single-connection forward until the peer closes.

**Fix.** Shut down the client socket when the copy thread finishes, and give `read_head` a deadline.

Koukyosyumei:
Fixed in #465.

PR #465, opened 2026-08-07
fix: bound the waits, threads and disk a box can drive unboundedly

Koukyosyumei:
Six lifecycle defects where h5i could be wedged or a host resource exhausted, with the box choosing when.

| Issue | Defect |
|---|---|
| #419 | Supervised egress run hung forever when the slirp uplink failed — the comment claimed a `poll()` timeout that did not exist |
| #420 | `EgressNetns::drop` joined the helper before closing the pipe it was blocked reading |
| #432 | Web forward hung and stopped serving when the box's stream closed first; `read_head` had no timeout |
| #435 | Auth proxy bounded each `read()` but not the head, so a slowloris starved the agent's own credentialed egress |
| #437 | Unbounded thread per proxy connection; uncapped tee-shim writes into the host spool |
| #444 | A named microVM survived an abnormal h5i exit with no backstop |

Notes on two of them:

**#437, the tee shim.** The shim is POSIX `sh` (`#!/.h5i/orig/bin/sh`), so process substitution is unavailable. The cap is a second fifo whose reader keeps the first N bytes and then drains the remainder to /dev/null — the drain is load-bearing, since without it `tee` takes EPIPE once the cap is hit and the command's passthrough output stops with it. Verified as a standalone shell script: 5500/6000 bytes passed through, 100/100 recorded at a 100-byte cap, exit code preserved, no hang.

**#444.** Reaping is keyed on the pid embedded in the sandbox name; `kill(pid, 0)` returning EPERM counts as alive, so a sandbox belonging to another user is never reaped.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` clean
- h5i-sandbox 297 passed, h5i-core 225 passed, console_api 8 passed
- env_integration 115 passed / 3 failed — the same three process-tier failures that fail on clean `main` on this host

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/h5i-core/src/view.rs
+++ b/crates/h5i-core/src/view.rs
@@ -856,12 +856,23 @@ impl Forward {
 
         // Out: frames to the human, unconditionally. Watching never collides,
         // so this direction has no policy at all.
+        //
+        // When the BOX's stream ends first (its browser exits, or `stream
+        // disable` runs), this copy returns and the client must be shut down
+        // too — the page has no way to learn the stream is gone, so it holds
+        // the socket open and `pump_input` below blocks on a peer that will
+        // never close. `serve()` takes one connection at a time, so that
+        // stopped the forward serving anything until the human closed the tab.
+        let client_shutdown = client.try_clone()?;
         let out = std::thread::spawn(move || {
-            std::io::copy(&mut upstream, &mut client_out).unwrap_or(0)
+            let n = std::io::copy(&mut upstream, &mut client_out).unwrap_or(0);
+            let _ = client_shutdown.shutdown(std::net::Shutdown::Read);
+            n
         });
         // In: gated by the control lock.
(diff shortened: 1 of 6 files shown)


slegarraga (approved):
Thanks for this focused lifecycle hardening. I checked the diff against the existing code and ran `cargo test -p h5i-sandbox --lib microvm::tests::` on this branch: 34 passed. The timeout and deadline changes are consistent with the existing per-read timeouts, and the tee cap approach preserves passthrough while bounding the spool.

One optional follow-up: `reap_orphaned_sandboxes` is the riskiest new path, and the tests cover `pid_alive` and marker lifecycle but not an actual sweep removing a dead marker or leaving a live one alone. A fake `bin` test would pin that behavior. Not blocking from my side.

PR #475, opened 2026-08-07
chore: remove unused root dependencies

### LINKED ISSUE #463: Root package declares chrono and serde but does not use them
`Cargo.toml:42,44`. No `chrono` or `serde::` references exist anywhere under `src/`; only `serde_json` is used by the bin crate.

Dead weight in the dependency graph of the primary package.

slegarraga:
Closes #463.

Removes chrono and serde from the root package because the workspace does not use them there. Verified with cargo check, clippy default, and clippy --no-default-features.


slegarraga:
Quick heads-up on the Linux check: the only failure is the existing `env::tests::locks_allow_one_writer_plus_many_observers_and_teardown_drains_observers` lock-teardown test, not anything touched by this dependency cleanup. The macOS build/test and all other checks pass. Happy to rebase or push a retry commit if that would help.

Koukyosyumei:
@slegarraga, thank you for your contribution! The changes look great to me. I’ll merge this once the `h5i box share` feature has been merged.

Koukyosyumei:
Reviewed this in a confined box (`h5i box --pr 475`) and pushed a merge of `main` that resolves the conflict. Summary of what I found and did.

## `chrono` was no longer safe to remove

The removal was correct when it was written, but `main` has moved ahead since. `9312cae8` ("h5i box share") added `src/cli/share.rs`, which calls `chrono::Utc::now()` in eight places (`src/cli/share.rs:169`, `:195`, `:222`, `:238`, `:544`, `:593`, `:753`, `:759`). That module is gated on `share-tunnel`, which is on by default via `share`, so taking the removal as-is would have broken the default build.

Git would not have caught this: `main` never touched the `chrono` line, so the deletion merged silently. The textual conflict was only over the adjacent `h5i-share` line. It is a semantic conflict, not a textual one.

`serde` checks out. No `serde::`, `Serialize` or `Deserialize` anywhere in `src/` or `tests/` on the merged tree, only the word in two doc comments. It stays removed.

## Resolution

- `Cargo.toml`: restored `chrono = { version = "0.4", features = ["serde"] }`, `serde` stays removed.
- `Cargo.lock`: took `main`'s version and dropped only `serde` from the `h5i` package entry. The `[[package]] serde` entry itself stays, since 46 other packages still depend on it.

The net diff against `main` is now exactly the two lines removing `serde`, which is what this PR should be. `cargo metadata --locked` and `cargo check --locked --workspace` both pass on the merged tree.

## Worth a separate change

In the spirit of the original PR: `chrono` is used only by `src/cli/share.rs`, so it could be `optional = true` and pulled in by `share-tunnel`, exactly as `tokio` already is a few lines below. That would keep a `--no-default-features` build from dragging chrono in. Left alone here because it is feature wiring rather than conflict resolution.

PR #489, opened 2026-08-11
fix: warn when H5I_AGENT is invalid

### LINKED ISSUE #458: An invalid $H5I_AGENT silently degrades to `human`
src/cli/boxes.rs:365-375. A value containing a space or a dot — for example `H5I_AGENT="claude code"` — fails the identity filter, and the box is created under the `human` identity and branch namespace with no warning.

That mislabels provenance and changes the auto-picked profile.

**Fix.** Emit a `warning:` line on the fallback so the user knows their identity was not used.

snowyukitty:
An invalid but present `H5I_AGENT` currently falls back to `human` without telling the user. Emit one warning on stderr when the value fails the existing identity grammar, while preserving the `human` fallback and successful create behavior.

Unset identities remain silent, valid values keep their current behavior, JSON stdout stays parseable, and the warning does not echo the invalid value. The integration test covers punctuation, whitespace, empty, overlong, non-ASCII, valid, trimmed, and unset identities. The manual now documents the validation and fallback behavior.

Closes #458

### Testing

- `cargo test --locked --no-default-features --test env_integration create_warns_for_invalid_agent_identity_without_contaminating_json -- --exact`
- `cargo test --locked --no-default-features --bin h5i`
- `cargo clippy --locked --no-default-features --bin h5i --test env_integration -- -D warnings`
- `cargo build --locked --no-default-features --bin h5i --test env_integration`
- `python scripts/gen_manual.py`

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -914,7 +914,7 @@ All optional; h5i ships with working defaults.
 
 | Variable | Purpose |
 |---|---|
-| `H5I_AGENT` | Agent identity used for the branch namespace and automatic runtime profile selection. Leading and trailing whitespace is trimmed; valid values contain 1–64 ASCII letters, digits, hyphens, or underscores. An unset value silently uses `human`; a present invalid value warns on stderr and falls back to `human`. |
+| `H5I_AGENT` | Which runtime a box is scoped to (`claude`, `codex`). Decides the env's branch namespace and the `agent` profile's credentials and egress. The namespace takes 1–64 ASCII letters, digits, hyphens, or underscores after trimming; unset is `human` silently, anything else warns on stderr and namespaces the box under `human`. |
 | `H5I_DEFAULT_ISOLATION` | Pin this clone's default tier when `--isolation` is not given. `--isolation auto` re-probes past it. |
 | `H5I_SECRET_<NAME>` | Default source for a secret grant `<NAME>`. Injected for one run, redacted from evidence, audited by fingerprint. |
 | `H5I_SKILL_DIR` | Where `h5i skill install` writes. |
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -914,7 +914,7 @@ All optional; h5i ships with working defaults.
 
 | Variable | Purpose |
 |---|---|
-| `H5I_AGENT` | Which runtime a box is scoped to (`claude`, `codex`). Decides the `agent` profile's credentials and egress. |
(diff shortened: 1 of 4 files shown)


Koukyosyumei (approved):
Approving. The behavior is right and the test earns its keep.

The core change is narrow and correct: `agent_identity()` now distinguishes "unset" from "present but unusable" instead of collapsing both through a `filter(...).unwrap_or(...)`, so the silent bare-shell case stays silent while a typo'd identity says so. `NotUnicode` routes to the same warning rather than being swallowed, which is the case the old `.ok()` dropped on the floor. Not echoing the rejected value is the right call for ambient input of unknown provenance, and the test pins that rather than trusting it.

The test table is the part I'd have asked for if it weren't already here: punctuation, embedded whitespace, empty, overlong, non-ASCII, valid, trimmed-valid, and unset, each asserting the manifest agent, the env id, the branch, and that stdout stays parseable JSON. That last assertion is what keeps the warning from becoming a breaking change for anything piping `--json`.

Three commits added on top while reviewing:

- `6f1f854d0` corrects the `H5I_AGENT` row in the manual. It claimed the fallback covers "the branch namespace and automatic runtime profile selection", but only the namespace falls back. `AgentRuntime::detect()` reads the raw var and `from_identity` matches `starts_with("codex")` on it, so `H5I_AGENT=codex.code` warns "using 'human'", namespaces under `human/`, and still hands the box Codex credentials and OpenAI egress. The row told the reader the opposite of what happens.
- `f33d3134e` fixes the CI break: `clippy 1.97` fires `collapsible_if` across the nested `if let Some(value)` / `if !value.is_empty()`, which is invisible on older clippy. Worth noting that clippy's own suggestion here is a trap: it proposes a let-chain, and CI compiles the merge commit (edition 2024, parses fine) while this branch is still edition 2021, where the same line is a hard parse error. `Option::filter` collapses it on both.

Two things for follow-up rather than blockers:

1. `agent_identity()` validates and falls back, but `AgentRuntime::detect()` bypasses it entirely and re-reads the raw var, so an invalid identity still selects a runtime from the unvalidated string. Unifying those would change behavior (`codex.code` would flip from Codex to Claude credentials), so it does not belong in a warning-only PR, but it is the actual split underneath the doc bug fixed above.
2. This branch is based on an edition-2021 main. It merges cleanly and CI tests the merge, so nothing is broken, but a rebase would make what you build locally match what CI checks.

PR #601, opened 2026-09-03
docs: Add x eget installation method

lunrenyi:
[x-cmd](https://www.x-cmd.com/) is a modern Shell toolkit that gives AI agents and developers powerful, portable, and composable command-line capabilities.

`x eget` allows users to install h5i directly from its GitHub release:

```sh
x eget use h5i-dev/h5i
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ h5i browser read https://example.com          # or: one page, no session
 curl -fsSL https://h5i.dev/install.sh | sh
 # curl -fsSL https://raw.githubusercontent.com/h5i-dev/h5i/main/install.sh | sh  # if you would rather not add a domain to the chain:
 # cargo install --path .                                                         # build from source
+# x eget use h5i-dev/h5i                                                         # If you are an x-cmd user.
 ```
 
 The agent-facing interface is a skill, and the binary carries it:


lunrenyi:
By the way, I'd also like to introduce [x-cmd/install](https://github.com/x-cmd/install).

It is an open-source community dedicated to collecting and maintaining more installation methods for more software.

If your project has an installation method that you'd like to make available to more users, you're welcome to contribute it to the community.

Once added, users can browse and search for available installation methods at [x-cmd.com/install](https://x-cmd.com/install/), or discover and install these tools directly through the `x install` command.

Project authors are also very welcome to contribute installation methods for their own projects and share them with the community!

Koukyosyumei:
Thanks for the contribution. We tested x eget against h5i’s current release assets and found that it may select an incompatible binary on unsupported architectures, such as the ARM64 macOS build on Intel macOS. It can also fall back to a third-party download mirror. Since listing this in the main installation section would imply official support, we cannot add it until platform matching fails safely and the downloaded artifact is always verified.

lunrenyi:
Thanks for testing this and for the detailed feedback!

That makes sense. The platform matching and artifact verification definitely need to be handled more carefully before presenting x eget as an official installation method.

I'll work on these issues and revisit this in a future PR. Thanks again for taking the time to test it!

